PR #8, opened 2020-09-19
Add note about --expt-relaxed-constexpr to readme

jlebar:
Note that you can in fact invoke constexpr functions even in the standard library from device code if you pass the right flag to nvcc.

Not sure if this change is welcome, but it seems relevant to the interests of people who might want to use this library but e.g. can't.

### DIFF
--- a/docs/readme.md
+++ b/docs/readme.md
@@ -31,7 +31,8 @@ When used with NVCC, NVIDIA C++ Standard Library facilities live in their own
   the host compiler's Standard Library:
 
 * `std::`/`<*>`: Your host compiler's Standard Library that works in
-      `__host__` code only.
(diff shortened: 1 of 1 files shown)


brycelelbach:
Sure, why not, it's useful info.

brycelelbach:
Ah, I should've put a "when used with NVCC" qualifier in there. We are trying to explicitly identify any behavior that is NVCC specific.

jlebar:
> Ah, I should've put a "when used with NVCC" qualifier in there. We are trying to explicitly identify any behavior that is NVCC specific.

+1 to that.

clang has the same behavior by default; disable it with `-fno-cuda-host-device-constexpr`

PR #14, opened 2020-09-21
Fix missing include in limits when using NVRTC

benbarsdell:
This is required to provide the definitions of `__CHAR_BIT__`, `__FLT_MANT_DIG__` etc.

### DIFF
--- a/include/cuda/std/limits
+++ b/include/cuda/std/limits
@@ -9,6 +9,9 @@
 #ifndef _CUDA_LIMITS
 #define _CUDA_LIMITS
 
+#ifdef __CUDACC_RTC__
+    #include "climits"
+#endif
 #include "type_traits"
 #include "version"
 


brycelelbach:
@griwes shouldn't QA have caught this? Do we have rest coverage for this?

griwes:
We have not done any QA on `<limits>`, we haven't put it on the list of officially supported headers yet.

Zenix27 (approved):

brycelelbach:
@griwes what are we doing with this PR, it's marked 1.3.0. Is it tested? Is it shipping?

griwes:
If it fixes the described problem then we should probably just ship it. I expect that something similar is what would ultimately need to be done when we start QAing limits. This passes an NVRTC run, so should be fine.

griwes (approved):

PR #17, opened 2020-09-22
`<tuple>` enabled on gcc and clang 

wmaxey:
Builds are clean on nvrtc, win32, and major gcc and clang versions.

MSVC is disabled.


brycelelbach:
@wmaxey let's ship `<tuple>` for non-MSVC compilers this week in 1.3.0 for the next CUDA toolkit.

wmaxey:
> @wmaxey let's ship `<tuple>` for non-MSVC compilers this week in 1.3.0 for the next CUDA toolkit.

@brycelelbach 
Sounds good.

I've noticed that the reference construction issue is present in gcc-7.5 similar to gcc-8. 

Tests are exempt on MSVC so it should be good to go there. Should we try to apply a workaround for these compilers?

Re: What we uncovered in August: there might be a few approaches that would satisfy NVCC's default reference construction complaint.

brycelelbach:
> I've noticed that the reference construction issue is present in gcc-7.5 similar to gcc-8.

Do we have a workaround for that?

brycelelbach:
> Re: What we uncovered in August: there might be a few approaches that would satisfy NVCC's default reference construction complaint.

Can we implement any of them this week?

wmaxey:
I've committed the changes that WAR the issue for gcc-8 and added a test to demonstrate if the fix is no longer required.

wmaxey:
@brycelelbach @griwes This is ready to be merged in. Just need a review to satisfy GH if I'm not mistaken.

libcxx/include/tuple
```diff
@@ -223,12 +224,13 @@ public:
...
-              class = _EnableIf<
+              _EnableIf<
                   _And<
                       _IsNotSame<__uncvref_t<_Tp>, __tuple_leaf>,
                       is_constructible<_Hp, _Tp>
-                    >::value
-                >
+                    >::value,
+                  bool
+                > = false
```
griwes, line 233: I'm really curious why this change was needed.

libcxx/include/tuple
```diff
@@ -939,10 +963,11 @@ public:
...
 
     template <class _Tuple,
-              class = typename enable_if
+              typename enable_if
                       <
-                         __tuple_assignable<_Tuple, tuple>::value
-                      >::type
+                         __tuple_assignable<_Tuple, tuple>::value,
+                         void*
+                      >::type = nullptr
```
griwes, line 970: A clever trick here is to not add the second argument, but add a `*` after `::type`. But this is fine too ;)

libcxx/include/tuple
```diff
@@ -1132,7 +1156,11 @@ struct __ignore_t
 };
 
 namespace {
+#ifdef __CUDA_ARCH__
+  _LIBCUDACXX_INLINE_VAR __constant__ __ignore_t<unsigned char> ignore = __ignore_t<unsigned char>();
+#else
   _LIBCUDACXX_INLINE_VAR constexpr __ignore_t<unsigned char> ignore = __ignore_t<unsigned char>();
+#endif
```
griwes, line 1163: A similar - though not identical - pattern also appeared in the complex changes; I wonder if there's a nice abstraction we could have for when we need to do selection of essentially a single token like this.

griwes (approved):
Minor comments, but don't gate shipping this on them.

libcxx/include/tuple
wmaxey: I've seen less SFINAE issues with MSVC when using value assignment rather than types. This may have been done out of an abundance of caution.

libcxx/include/tuple
wmaxey: I'll update this to an actual `_EnableIf` from the internal meta functions.

libcxx/include/tuple
wmaxey: I'll turn this into an issue. :)

PR #18, opened 2020-09-22
Docs: Simplified a few of the build samples.

wmaxey:
Also added a warning about my misadventures when compiling with MSVC.


brycelelbach (approved):
Looks good to me.

PR #21, opened 2020-09-23
Disable __builtin_is_constant_evaluated with NVCC in C++11 mode

wmaxey:
NVCC reports that uses of `__builtin_is_constant_evaluated` in a constexpr evaluation are invalid:

`detail/libcxx/include/type_traits(4835): error #2416-D: constexpr function return is non-constant`

The fix for now is a WAR that removes the use of the builtin when compiling under NVCC:

I've added a test that should let us know when it's functioning again.

### DIFF
--- a/libcxx/include/__config
+++ b/libcxx/include/__config
@@ -643,7 +643,8 @@ typedef __char32_t char32_t;
 #define _LIBCUDACXX_ATOMIC_ALWAYS_LOCK_FREE(...) __atomic_always_lock_free(__VA_ARGS__)
 #endif
 
-#if __has_builtin(__builtin_is_constant_evaluated)
+// NVCC in C++11 mode freaks out about `__builtin_is_constant_evaluated`.
+#if __has_builtin(__builtin_is_constant_evaluated) && !(defined(_LIBCUDACXX_COMPILER_NVCC) && _LIBCUDACXX_STD_VER < 14)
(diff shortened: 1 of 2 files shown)


libcxx/include/__config
```diff
@@ -647,6 +647,14 @@ typedef __char32_t char32_t;
 #define _LIBCUDACXX_IS_CONSTANT_EVALUATED(...) __builtin_is_constant_evaluated(__VA_ARGS__)
 #endif
 
+// TODO: Builtin is not constexpr, this can be tested with an integral_constant
+//       but _only_ in C++11 mode. (fix for NVCC+clang-9 and up)
+#if defined(__CUDACC__)
+#if __cplusplus < 201402L
+#undef _LIBCUDACXX_IS_CONSTANT_EVALUATED
+#endif
+#endif
```
brycelelbach, line 656: Can we do this without an `#undef` by moving it into the NVCC configuration section, and only defining `_LIBCUDACXX_IS_CONSTANT_EVALUATED` if not Clang 9 and not C++11?

brycelelbach (approved):

libcxx/include/__config
brycelelbach: Also I'm confused, does this affect GCC too? If so we should have the comment say that. The test indicates it affects GCC 9 and 10 and Clang 9 and 10?

griwes (approved):

PR #32, opened 2020-10-02
`<complex>` with `long double` disabled

wmaxey:
Adds support for `<complex>` with most types. `

long double`, however, is disabled due to platform support differences.


.upstream-tests/maintenance/stdlib-headers
```diff
@@ -6,6 +6,7 @@
 <bit>
 <bitset>
 <cassert>
+<ccomplex>
```
griwes, line 9: This should be `<complex>`.

libcxx/include/complex
```diff
@@ -105,32 +103,14 @@ public:
...
-    complex<long double>& operator/=(long double);
-
-    template<class X> complex<long double>& operator= (const complex<X>&);
-    template<class X> complex<long double>& operator+=(const complex<X>&);
-    template<class X> complex<long double>& operator-=(const complex<X>&);
-    template<class X> complex<long double>& operator*=(const complex<X>&);
-    template<class X> complex<long double>& operator/=(const complex<X>&);
+
+
 };
```
griwes, line 114: This diff looks weird; shouldn't the whole block just be gone? (I get that it's in the documentation comment at the top, it';s still weird.)

libcxx/include/complex
```diff
@@ -438,93 +410,29 @@ public:
...
-        {
-            *this = *this / complex(__c.real(), __c.imag());
-            return *this;
-        }
+    template <typename _Dummy = void>
+    _LIBCUDACXX_INLINE_VISIBILITY _LIBCUDACXX_CONSTEXPR complex(long double __re = 0.0, long double __im = 0.0)
+        {static_assert(is_same<_Dummy, void>::value, "complex<long double> is not supported");}
+    template <typename _Tp>
+    _LIBCUDACXX_INLINE_VISIBILITY _LIBCUDACXX_CONSTEXPR complex(const complex<_Tp> &__c)
+        {static_assert(sizeof(_Tp) == 0, "complex<long double> is not supported");}
```
griwes, line 423: `sizeof(_Tp) == 0` is technically always true, so this function template can produce no valid template instantiation, therefore an overeager compiler could just fire the static assertion at parsing time (template definitions that have no valid instantiations are ill-formed, NDR). Repeat the `_Dummy` trick from above?

libcxx/include/complex
```diff
@@ -892,7 +790,7 @@ typename enable_if
 >::type
 arg(_Tp __re)
 {
-    return atan2(0., __re);
+    return atan2(0., static_cast<double>(__re));
```
griwes, line 793: Why is this needed?

libcxx/include/complex
```diff
@@ -1456,24 +1354,13 @@ operator<<(basic_ostream<_CharT, _Traits>& __os, const complex<_Tp>& __x)
 
 #endif // __cuda_std__
 
-#if _LIBCUDACXX_STD_VER > 11
+#if _LIBCUDACXX_STD_VER > 11 && defined(_LIBCUDACXX_HAS_STL_LITERALS)
```
griwes, line 1357: Why are we disabling these for `cuda::std::complex`?

griwes (changes requested):
One overarching general comment: striking the long double code from the upstream headers like this will break testing of the embedded libc++, and may make shipping this in a certain form in the future. For the time being I think the best way to avoid that is to wrap the removed code blocks into `#ifndef __cuda_std__ / #endif`, and the added code blocks into `#ifdef __cuda_std__`.

libcxx/include/complex
griwes: Actually remembering my overarching comment now, maybe also wrap it with ifdefs inside this comment ;)

.upstream-tests/maintenance/stdlib-headers
griwes: Ah! I get it now. Disregard ;)

libcxx/include/complex
wmaxey: Caught me there, thanks. I just did a line by line removal and forgot about the specialization.

libcxx/include/complex
wmaxey: Will do.

libcxx/include/complex
wmaxey: STL literals are reserved. UDLs seem to be enforced to have an `_prefix`.

libcxx/include/complex
wmaxey: This is an error from when long double was shunted through `complex<double>`  in order to satisfy math intrinsics on device.

libcxx/include/complex
griwes: Ah, and so we trigger warnings when we try to define them outside of `std::`?

wmaxey: Yeah, NVCC picks them up first so I'm not sure if we can do anything `#pragma` related.

libcxx/include/complex
wmaxey: Well, apparently that wasn't the cause. There's no integral `atan2` in device code and I think the signature prevents promotion to double somehow. Do you have advice on a solution? I can split the integral path from double and make it more explicit that way.

libcxx/include/complex
```diff
@@ -708,13 +708,14 @@ operator/(const complex<_Tp>& __z, const complex<_Tp>& __w)
     if (__libcpp_isfinite_or_builtin(__logbw))
     {
         __ilogbw = static_cast<int>(__logbw);
-        // ldexp is equivalent to scalbn when FLT_RADIX == 2
-        __c = ldexp(__c, -__ilogbw);
-        __d = ldexp(__d, -__ilogbw);
+        // Promote to double as scalbn is not generic
```
wmaxey, line 711: @griwes I'm testing this on Windows now, but I want to get your opinion first. In lieu of implementing a function to correctly dispatch to `scalbnf/scalbnl/scalbn` I though this seemed fine.

wmaxey:
Virtual 29333392.1 has one failure that was fixed after an amend; no need for resubmit.

griwes (approved):

libcxx/include/__config
```diff
@@ -209,6 +209,11 @@
 // This is not mutually exclusive with other compilers, as NVCC uses a host
 // compiler.
 #  define _LIBCUDACXX_COMPILER_NVCC
+#else
+// As NVCC does not really support `long double` the type is ignored
+#  define _LIBCUDACXX_HAS_COMPLEX_LONG_DOUBLE
+// Also, since we cannot silence non '_' prefixed literal warnings, we need to disable some UDLs
+#  define _LIBCUDACXX_HAS_STL_LITERALS
```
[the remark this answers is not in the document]
wmaxey: Ah, so instead we predicate on `_LIBCUDACXX_NO_COMPLEX_LONG_DOUBLE`?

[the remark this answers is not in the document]
griwes: No, I _think_ the macros are fine here; it's the comments that seem somewhat unclear, since they talk about NVCC in the non-NVCC branch.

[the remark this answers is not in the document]
wmaxey: I went for not changing the documentation, but I understand. I can make that change without re-running CI.

PR #42, opened 2020-10-07
Update broken_constant_eval.pass.cpp

### LINKED ISSUE #41: Remove `gcc-9` as expected failure from `broken_constant_eval.pass.cpp`
### gcc-9/10 non-constant pass/fail

* `pass` - `gcc-8.4.0`
* `pass` - `gcc-8.4.1`
* `untested` - `gcc-9.0.0`
* `pass` - `gcc-9.1.0`
* `pass` - `gcc-9.3.1`
* `pass` - `gcc-10.0.0`
* `fail` - `gcc-10.1.0`
* `fail` - `gcc-10.2.0`

This broke tests covering gcc-9, and will break tests with gcc-10.0.
This seems consistent with different versions of NVCC.

#21 is root cause.

wmaxey:
This can fix #41
However, I don't know if there is a way to more specifically target a minor gcc version.

### DIFF
--- a/.upstream-tests/test/std/utilities/utilities.general/broken_constant_eval.pass.cpp
+++ b/.upstream-tests/test/std/utilities/utilities.general/broken_constant_eval.pass.cpp
@@ -11,7 +11,7 @@
 // UNSUPPORTED: windows, icc, pgi
 // XFAIL: clang-9 && c++11
 // XFAIL: clang-10 && c++11
-// XFAIL: gcc-9 && c++11
+// gcc 10.0 is expected to pass, but later versions do not.
 // XFAIL: gcc-10 && c++11
 
 


.upstream-tests/test/std/utilities/utilities.general/broken_constant_eval.pass.cpp
```diff
@@ -11,7 +11,7 @@
 // UNSUPPORTED: windows, icc, pgi
 // XFAIL: clang-9 && c++11
 // XFAIL: clang-10 && c++11
-// XFAIL: gcc-9 && c++11
+// gcc 10.0 is expected to pass, but later versions do not.
```
griwes, line 14: Is this meant to say GCC 9?

griwes (approved):

.upstream-tests/test/std/utilities/utilities.general/broken_constant_eval.pass.cpp
wmaxey: See the list in the issue. I might have an inverted sense of what counts as passing because it's an expected failure.

Regardless, 10.0 is not susceptible while 10.1 and 10.2 are.

griwes: I see. 10.0 is not a released version, so to avoid confusion we should probably not refer to it.

PR #43, opened 2020-10-07
Decouple Atomic from host on MSVC

wmaxey:


griwes:
Looks good overall pending the first review comment; please ping me when you un-WIP this.

include/cuda/std/detail/__atomic
```diff
@@ -190,46 +200,42 @@ struct __cxx_atomic_base_impl_default {
...
-    ::std::atomic_init(&__a->__a_value, __cxx_atomic_alignment_wrap(__val));
-#endif
 }
 template<class _Tp, int _Sco>
 __host__ __device__ inline void __cxx_atomic_store(__cxx_atomic_base_impl_default<_Tp, _Sco> volatile* __a, _Tp __val, int __order) {
 #ifdef __CUDA_ARCH__
     detail::__atomic_store_n_cuda(&__a->__a_value, __cxx_atomic_alignment_wrap(__val), __order, detail::__scope_tag<_Sco>());
 #else
-    ::std::atomic_store_explicit(&__a->__a_value, __cxx_atomic_alignment_wrap(__val), (::std::memory_order)__order);
+    __atomic_store(&__a->__a_value, &__val, __order);
```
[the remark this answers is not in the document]
wmaxey: Great question, let me recheck why I had made this change. It was done to fix a build.

include/cuda/std/detail/__atomic
```diff
@@ -190,46 +200,42 @@ struct __cxx_atomic_base_impl_default {
...
 #endif
 }
 template<class _Tp, int _Sco>
 __host__ __device__ inline _Tp __cxx_atomic_load(__cxx_atomic_base_impl_default<_Tp, _Sco> const volatile* __a, int __order) {
 #ifdef __CUDA_ARCH__
     return __cxx_atomic_alignment_unwrap(detail::__atomic_load_n_cuda(&__a->__a_value, __order, detail::__scope_tag<_Sco>()));
 #else
-    return __cxx_atomic_alignment_unwrap(::std::atomic_load_explicit(&__a->__a_value, (::std::memory_order)__order));
+    alignas(_Tp) unsigned char __buf[sizeof(_Tp)];
```
wmaxey, line 224: @griwes it looks like I tore off another alignment unwrap here.

Are these `char*` output buffers a code smell? This was done to fix initialization warnings.

wmaxey:
Testing on a GV100 looks good. No failures with 1 unsupported test. I'll kick off a CI run now.

I've replaced the `__cxx_atomic_alignment_unwrap`/`wrap` and in the case of some functions like [`__cxx_atomic_compare_exchange_strong`](https://github.com/NVIDIA/libcudacxx/commit/03a79ef43a461065683f598897477b527b0b8bd2#diff-16758c8c5140da43455f3a53ee379490238cea6413ee9ed3ebb6b6e97afcb936R249) I've removed them because the underlying atomic handles it.

I'm unsure if that is the correct thing to do however.

include/cuda/std/detail/__atomic_derived
```diff
@@ -43,23 +43,23 @@ template<class _Type, class _Delta, class _Scope, typename _CUDA_VSTD::enable_if
 _Type __device__ __atomic_fetch_add_cuda(_Type volatile *__ptr, _Delta __val, int __memorder, _Scope __s) {
 
     _Type __expected = __atomic_load_n_cuda(__ptr, __ATOMIC_RELAXED, __s);
-    _Type const __desired = __expected + __val;
+    _Type __desired = __expected + __val;
     while(!__atomic_compare_exchange_cuda(__ptr, &__expected, &__desired, true, __memorder, __memorder, __s))
-        ;
+        __desired = __expected + __val;
```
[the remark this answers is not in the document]
ogiroux: Yes

include/cuda/std/detail/__atomic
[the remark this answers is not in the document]
ogiroux: I think this needs to reintroduce the wrap.

include/cuda/std/detail/__atomic
ogiroux: Why not use __atomic_load_n here instead?

include/cuda/std/detail/__atomic
```diff
@@ -259,55 +267,55 @@ __host__ __device__ inline _Tp __cxx_atomic_fetch_add(__cxx_atomic_base_impl_def
...
+    return __atomic_fetch_add(&__a->__a_value, __delta, __order);
 #endif
 }
 template<class _Tp, int _Sco>
 __host__ __device__ inline _Tp* __cxx_atomic_fetch_add(__cxx_atomic_base_impl_default<_Tp*, _Sco> volatile* __a, ptrdiff_t __delta, int __order) {
 #ifdef __CUDA_ARCH__
     return detail::__atomic_fetch_add_cuda(&__a->__a_value, __delta, __order, detail::__scope_tag<_Sco>());
 #else
-    return ::std::atomic_fetch_add_explicit(&__a->__a_value, __delta, (::std::memory_order)__order);
+    return __atomic_fetch_add(&__a->__a_value, __delta * __skip_amt<_Tp*>::value, __order);
```
ogiroux, line 278: It doesn't seem right to me that we should need the skip amount in this layer. The layer below should be doing that.

ogiroux: Basically, every deviation between the CUDA_ARCH side and this side looks like a bug to me.

include/cuda/std/detail/__atomic
wmaxey: __atomic_load_n does not work for non-integral non-pointer types: https://gcc.godbolt.org/z/osrcs1

Maybe I'm mistaken in some way about its usage?

include/cuda/std/detail/__atomic
wmaxey: > It doesn't seem right to me that we should need the skip amount in this layer. The layer below should be doing that.

I'd agree, but there is no layer at the compiler intrinsic level for GCC. At that point incrementing by the sizeof(_Tp) is necessary. https://github.com/NVIDIA/libcudacxx/blob/main/libcxx/include/atomic#L846

> Basically, every deviation between the CUDA_ARCH side and this side looks like a bug to me.

@griwes, @jrhemstad, and I had a meeting today about how we could resolve some of this with a better platform layering framework. There's some neat ideas on the table for making this nesting doll thing be a bit cleaner.

It would be relevant to know what things are being done wrong ahead of time.

ogiroux (approved):
I think this looks good.

wmaxey:
Atomic decouple builds clean on CI.

SC: 29322243.2

PR #47, opened 2020-10-20
Docs: Shard metadata, copy-edit synchronization library docs

brycelelbach:
- Shard the Jekyll metadata from `_config.yml` into each of the individual
  markdown files.
- Copy-edit, reformat, and add links to synchronization library documentation.
- Fill in docs for `<utility>`.
- Remove development model section until we get around to it.

### DIFF
--- a/docs/api/synchronization_library/atomic.md
+++ b/docs/api/synchronization_library/atomic.md
@@ -1,56 +1,75 @@
+---
+grand_parent: API
+parent: Synchronization Library
+nav_order: 1
+---
+
 # `<cuda/std/atomic>`, `<cuda/atomic>`
 
-## Extensions in the `cuda::` namespace
+## Extensions
 
-The class template `atomic` takes an additional scope argument, defaulted to `thread_scope_system`.
+The class template `atomic` takes an additional [thread scope] argument,
+  defaulted to `thread_scope_system`.
 
 ```c++
-// This object is atomic for all threads in the system
-cuda::atomic<int> a; 
+// This atomic is suitable for all threads in the system.
+cuda::atomic<int, cuda::thread_scope_system> a;
 
-// This object has the same type as the previous
-cuda::atomic<int, cuda::thread_scope_system> b; 
+// This atomic has the same type as the previous one (`a`).
+cuda::atomic<int> b;
 
-// This object is atomic for threads in the same thread block
-cuda::atomic<int, cuda::thread_scope_block> c; 
+// This atomic is suitable for threads in the same thread block.
+cuda::atomic<int, cuda::thread_scope_block> c;
 ```
 
-The `atomic` class template specializations for integral and pointer types are extended with members `fetch_min` and `fetch_max`. These conform to the requirements in `[atomics.types.int]` and `[atomics.types.pointer]`, with keys _min_ and _max_, and operations `min` and `max`, respectively.
+The `atomic` class template specializations for integral and pointer types are
+extended with members `fetch_min` and `fetch_max`.
+These conform to the requirements in section [atomics.types.int] and
+  [atomics.types.pointer] of ISO/IEC IS 14882 (the C++ Standard),
+  with keys _min_ and _max_, and operations `min` and `max`, respectively.
 
-Note that conformance to these requirements include implicit conversions to unsigned types, if applicable.
+Note that conformance to these requirements include implicit conversions to
+  unsigned types, if applicable.
 
 ```c++
-cuda::atomic<int> a(1); 
-auto x = a.fetch_min(0); // operates as if unsigned
+cuda::atomic<int> a(1);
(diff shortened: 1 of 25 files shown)


docs/api/synchronization_library/atomic.md
```diff
@@ -1,56 +1,75 @@
+---
+grand_parent: API
+parent: Synchronization Library
+nav_order: 1
+---
+
 # `<cuda/std/atomic>`, `<cuda/atomic>`
 
-## Extensions in the `cuda::` namespace
```
griwes, line 3: Why are we dropping this part of the heading? The prose below doesn't seem to make it explicit that this only applies in `cuda::`, and the examples only hint at it.

docs/api/synchronization_library/barrier.md
```diff
@@ -1,29 +1,41 @@
...
 ```c++
-__shared__ cuda::barrier<cuda::thread_scope_block> b; // shared memory does not allow initialization
+// Shared memory does not allow initialization.
+__shared__ cuda::barrier<cuda::thread_scope_block> b;
 
-init(&b, 1); // use this free function to initialize this object
+init(&b, 1); // Use this friend function to initialize the object.
 ```
 
-In the `device::` sub-namespace a `__device__` free function is available that provides direct access to the underlying PTX state of a `barrier` object, if its scope is `thread_scope_block` and it is allocated in shared memory.
```
griwes, line 26: Qualify `device::` as `cuda::device::` if you want to avoid saying "subnamespace".

docs/api/synchronization_library/barrier.md
```diff
@@ -33,8 +45,9 @@ __device__ std::uint64_t* barrier_native_handle(
 
 }}
 ```
-* Expects: `b` is in shared memory;
-* Returns: a pointer to the PTX "mbarrier" subobject of the `barrier` object.
+
+- Expects: `b` is in `__shared__` memory.
+- Returns: a pointer to the PTX "mbarrier" subobject of the `barrier` object.
```
griwes, line 51: "Shared memory" is a phrase with meaning though, right? I think I prefer prose here instead of code font "`__shared__`".

docs/api/synchronization_library/barrier.md
```diff
@@ -44,34 +57,48 @@ auto ptr = barrier_native_handle(b);
...
+- it is in unified memory and the [`concurrentManagedAccess` property] is 1, or
+- it is in CPU memory and the [`hostNativeAtomicSupported` property] is 1.
 
-(Note, for objects of scopes other than `thread_scope_system` this is a data-race, and thefore also prohibited regardless of memory characteristics.)
+Note, for objects of scopes other than `thread_scope_system` this is a
+  data-race, and thefore also prohibited regardless of memory characteristics.
 
-Under Compute Capability 8 or above, when an object of type `barrier<thread_scope_block>` is placed in `__shared__` memory, the member function `arrive` performs a reduction of the arrival count among [coalesced](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#coalesced-group-cg) threads followed by the arrival operation in one thread. Programs shall ensure that this transformation would not introduce errors, for example relative to the requirements of `[thread.barrier.class]` paragraph 12 in N4860.
+Under CUDA Compute Capability 8 (Ampere) or above, when an object of type
+  `barrier<thread_scope_block>` is placed in `__shared__` memory, the member
```
griwes, line 75: Ditto about "`__shared__`" here.

griwes (approved):
Approving pending the comments below being addressed one way or the other.

docs/api/synchronization_library/atomic.md
brycelelbach: I just don't like it being so prosey. All the other section titles are short and succinct.

docs/api/synchronization_library/barrier.md
brycelelbach: It's a bad phrase anyways, blame @ogiroux.

docs/api/synchronization_library/barrier.md
brycelelbach: I don't want to say "shared memory" because people who are NOT familiar with CUDA will read that as... well... the traditional meaning of shared memory: e.g. interprocess shared memory.

docs/api/synchronization_library/barrier.md
brycelelbach: I don't want to say "shared memory" because people who are NOT familiar with CUDA will read that as... well... the traditional meaning of shared memory: e.g. interprocess shared memory.

docs/api/synchronization_library/atomic.md
griwes: It needs to go into the actual prose then.

PR #52, opened 2020-10-22
WIP: Implement atomic_ref

wmaxey:
Early draft of some minimum set of changes needed for `atomic_ref`

Currently working or done:
* [x] `atomic<>`
* [ ] `atomic_ref<>`
* [x] atomic tests
* [] `atomic_ref` tests


include/cuda/std/detail/__atomic
```diff
@@ -195,6 +192,126 @@ __host__ __device__ auto __cxx_atomic_alignment_unwrap(_Tp __value)
     return __cxx_atomic_alignment_unwrap(__value, integral_constant<bool, _LIBCUDACXX_ALIGNOF(_Tp) == sizeof(_Tp)>{});
 }
 
+template <typename _Tp>
+__host__ __device__ constexpr ptrdiff_t __cxx_atomic_skip(ptrdiff_t __value) {
+    using __t = decltype(*__cxx_atomic_alignment_unwrap(_Tp{}, integral_constant<bool, _LIBCUDACXX_ALIGNOF(_Tp) == sizeof(_Tp)>{}));
```
wmaxey, line 197: This invokes unwrap to get the underlying type (if needed) for incrementing pointers

eg: 
 * `int* == 4*increment`
 * `long* == 8*increment`
 * `custom_class* == sizeof(custom_class)*increment`

griwes: I wonder if instead we should just expose a `value_type` in the alignment wrapper and query that here.

include/cuda/std/detail/__atomic
```diff
@@ -185,130 +192,216 @@ __host__ __device__ auto __cxx_atomic_alignment_unwrap(_Tp __value)
...
+    constexpr size_t __skip = sizeof(__t);
+    return __skip * __value;
+}
+
+template <class _Tp>
+using __atomic_strip_cv_t = typename cuda::std::remove_cv<_Tp>::type;
+
+template <class _Tp, typename _Tag>
+__host__ __device__ inline void __raw_atomic_store(_Tp* __a, __atomic_strip_cv_t<_Tp> __val, int __order, _Tag) {
```
griwes, line 206: Flip the name parts maybe? I.e. `__atomic_raw_store`? This is not a strong opinion, but I _think_ I'd prefer that (OTOH I do recognize that the backend goes `__cxx_atomic` and not `__atomic_cxx` on names, so really either is fine I think - especially since this is further muddled by the CUDA backend having `_cuda` and the **very end**.)

libcxx/include/atomic
```diff
@@ -705,6 +719,72 @@ _LIBCUDACXX_INLINE_VISIBILITY inline _LIBCUDACXX_CONSTEXPR int __to_gcc_failure_
               __ATOMIC_CONSUME))));
 }
 
+template<class _Tp>
+_LIBCUDACXX_INLINE_VISIBILITY inline void __raw_atomic_store(_Tp* __a, _Tp __val, int __order) {
```
griwes, line 723: I'm trying to understand the layering here. From reading the files above (and I may revisit this when I read the changes below), it seemed to me that the `__raw_atomic_*` layer was to implement the C++ atomics backed that we plug into libc++; but now I'm also seeing it here, and the code is actually really similar to what I've seen above, only without scopes. This seems like repetition; can you explain why we need to duplicate some of these function shapes? (`__raw_atomic_store` seems fine, but the logic for `__raw_atomic_load` for instance is duplicated in our `__cxx_atomic_load` and that is somewhat concerning to me.)

libcxx/include/atomic
```diff
@@ -1939,6 +2019,318 @@ struct atomic<_Tp*>
...
+    operator _Tp() const noexcept { return this->load(); }
+
+    _LIBCUDACXX_INLINE_VISIBILITY
+    bool is_lock_free() const noexcept {
+        return __cxx_atomic_is_lock_free(sizeof(_Tp));
+    }
+
+    _LIBCUDACXX_INLINE_VISIBILITY
+    void store(_Tp __t, memory_order __m = memory_order_seq_cst) const noexcept {
+        __raw_atomic_store(__ptr, __t, __m, detail::__scope_tag<_Sco>());
```
griwes, line 2086: So looking at this (continuing from the previous comment) - does it mean that there's now two layers that the implementer of an atomic backend needs to provide? The `__cxx` layer and the `__raw` layer? Is there a way to require only one of them to be provided?

What I'm thinking here is that there may be a way to just provide a generic `__cxx` layer built on top of `__raw`, call it a new libc++ atomic backend, and have us just provide the `__raw` layer (and maybe the `__cxx` layer types - we could have the definitions of the types and the definitions of the generic layer on top of `__raw` be hidden behind different macros). Then, libcu++'s changes would be:
1) scopes added to all functions and types in this header;
2) types for the `__cxx` layers; and
3) the `__raw` layer.

PR #53, opened 2020-10-22
Add extended API documentation

c0riolis:
- [x] Do the required Jekyll changes
- [x] Fix relative links
- [x] Document `Group` concept & `aligned_size_t`
- [x] Add missing examples
- [x] Verify that all examples are compiling

The rendered version of this PR is available here:
https://draft.dll.pe/

### DIFF
--- a/docs/extended_api/asynchronous_operations_library/aligned_size_t.md
+++ b/docs/extended_api/asynchronous_operations_library/aligned_size_t.md
@@ -1,0 +1,61 @@
+---
+grand_parent: Extended API
+parent: Asynchronous operations library
+---
+
+# cuda::**aligned_size_t**
+
+Defined in header [`<cuda/barrier>`](../headers/barrier.md)
+
+Defined in header [`<cuda/pipeline>`](../headers/pipeline.md)
+
+```c++
+template<size_t Alignment>
+struct aligned_size_t;
+```
+
+The class template `cuda::aligned_size_t` is a _shape_ representing an extent of bytes with a statically defined (address and size) alignment.
+
+## Template parameters
+
+| Alignment | the address & size alignement of the byte extent |
+
+## Data members
+
+| [align](./aligned_size_t/align.md) | the alignment of the byte extent |
+| [value](./aligned_size_t/value.md) | the size of the byte extent      |
+
+## Member functions
+
+| [(constructor)](./aligned_size_t/constructor.md) | constructs an _aligned size_                                                      |
+| (destructor) [implicitly declared]               | trivial implicit destructor                                                       |
+| operator= [implicitly declared]                  | trivial implicit copy/move assignment                                             |
+| operator size_t                                  | implicit conversion to [`size_t`](https://en.cppreference.com/w/cpp/types/size_t) |
+
+## Notes
+
+If `value` is not a multiple of `align` the behavior is undefined.
+
+If `Alignment` is not a [valid alignment](https://en.cppreference.com/w/c/language/object#Alignment) the behavior is undefined.
+
+## Example
+
+```c++
+#include <cuda/barrier>
+
+__global__ void example_kernel(void * dst, void * src, size_t size)
+{
+    cuda::barrier<cuda::thread_scope_system> barrier;
+    init(&barrier, 1);
+
+    // Implementation cannot make assumptions about alignment
+    cuda::memcpy_async(dst, src, size, barrier);
+
+    // Implementation can assume that dst, src and size are 16-bytes aligned and may optimize accordingly
+    cuda::memcpy_async(dst, src, cuda::aligned_size_t<16>(size), barrier);
+
+    barrier.arrive_and_wait();
+}
+```
+
+[See it on Godbolt](https://godbolt.org/z/v7Ev9E){: .btn }
--- a/docs/extended_api/asynchronous_operations_library/memcpy_async.md
+++ b/docs/extended_api/asynchronous_operations_library/memcpy_async.md
@@ -1,0 +1,64 @@
+---
+grand_parent: Extended API
+parent: Asynchronous operations library
+---
+
+# cuda::**memcpy_async**
+
+Defined in header [`<cuda/barrier>`](../../api/synchronization_library/barrier.md)
+
+```c++
+template<typename Shape, thread_scope Scope>
+void memcpy_async(void * destination, void const * source, Shape size, barrier<Scope> & barrier);                        // (1)
+
+template<typename Group, typename Shape, thread_scope Scope>
+void memcpy_async(Group const & group, void * destination, void const * source, Shape size, barrier<Scope> & barrier);   // (2)
+```
+
+Defined in header [`<cuda/pipeline>`](../headers/pipeline.md)
+
+```c++
+template<typename Shape, thread_scope Scope>
+void memcpy_async(void * destination, void const * source, Shape size, pipeline<Scope> & pipeline);                      // (3)
+
+template<typename Group, typename Shape, thread_scope Scope>
+void memcpy_async(Group const & group, void * destination, void const * source, Shape size, pipeline<Scope> & pipeline); // (4)
+```
+
+Asynchronously copies `size` bytes from the memory location pointed to by `source` to the memory location pointed to by `destination`.
+Both objects are reinterpreted as arrays of `unsigned char`.
+
+`cuda::memcpy_async` have similar constraints to [`std::memcpy`](https://en.cppreference.com/w/cpp/string/byte/memcpy), namely:
+* If the objects overlap, the behavior is undefined.
+* If either `destination` or `source` is an invalid or null pointer, the behavior is undefined (even if `count` is zero).
+* If the objects are [potentially-overlapping](https://en.cppreference.com/w/cpp/language/object#Subobjects) the behavior is undefined.
+* If the objects are not of [`TriviallyCopyable`](https://en.cppreference.com/w/cpp/named_req/TriviallyCopyable) type the program is ill-formed, no diagnostic required.
+
+If _Shape_ is [`cuda::aligned_size_t`](./aligned_size_t.md)), `source` and `destination` are both required to be aligned on [`cuda::aligned_size_t::align`](./aligned_size_t/align.md), else the behavior is undefined.
+
+If `pipeline` is in a _quitted state_ (see [`pipeline::quit`](../synchronization_library/pipeline/quit.md)), the behavior is undefined.
+
+1. Binds the asynchronous copy completion to `barrier` and issues the copy in the current thread.
+2. Binds the asynchronous copy completion to `barrier` and cooperatively issues the copy across all threads in `group`.
+3. Binds the asynchronous copy completion to `pipeline` and issues the copy in the current thread
+4. Binds the asynchronous copy completion to `pipeline` and cooperatively issues the copy across all threads in `group`.
+
+## Template parameters
+
+| Group | a type satisfying the [_group concept_](../concepts/group.md)                                                                                                                  |
+| Shape | a type satisfying the [_shape concept_](../concepts/shape.md) (see [`size_t`](https://en.cppreference.com/w/c/types/size_t) and [`cuda::aligned_size_t`](./aligned_size_t.md)) |
+
+## Parameters
+
+| group       | the group of threads                                    |
+| destination | pointer to the memory location to copy to               |
+| source      | pointer to the memory location to copy from             |
+| size        | the number of bytes to copy                             |
+| barrier     | the barrier object used to wait on the copy completion  |
+| pipeline    | the pipeline object used to wait on the copy completion |
+
+## Example
+
+```c++
+TODO
+```
--- a/docs/extended_api/concepts/group.md
+++ b/docs/extended_api/concepts/group.md
@@ -1,0 +1,46 @@
+---
+grand_parent: Extended API
+parent: Concepts
+---
+
+# Group
+
+```c++
+struct Group {
+  static constexpr cuda::thread_scope thread_scope;
+  integral size() const;
+  integral thread_rank() const;
+  void sync() const;
+};
+```
+
+The _Group concept_ defines the requirements of a type that represents a group of cooperating threads.
+
+## Data members
+
+| thread_scope | the scope at which `Group::sync()` synchronizes memory operations and thread execution |
+
+## Member functions
+
+| size        | returns the number of participating threads                                                        |
+| thread_rank | returns a unique value for each participating thread (`0 <= Group::thread_rank() < Group::size()`) |
+| sync        | synchronizes the participating threads                                                             |
+
+## Notes
+
+This concept is defined for documentation purposes but is not materialized in the library.
+
+## Example
+
+```c++
+#include <cuda/atomic>
+
+struct single_thread_group {
+    static constexpr cuda::thread_scope thread_scope = cuda::thread_scope::thread_scope_thread;
+    size_t size() const { return 1; }
+    size_t thread_rank() const { return 0; }
+    void sync() const { }
+};
+```
+
+[See it on Godbolt](https://godbolt.org/z/453r3s){: .btn }
--- a/docs/extended_api/concepts/shape.md
+++ b/docs/extended_api/concepts/shape.md
@@ -1,0 +1,37 @@
+---
+grand_parent: Extended API
+parent: Concepts
+---
+
+# Shape
+
+```c++
+struct Shape {
+    operator size_t() const;
+};
+```
+
+The _Shape concept_ defines the requirements of a type that represents a byte extent with a particular memory layout.
+
+## Member functions
+
+| operator size_t | implicit conversion to [`size_t`](https://en.cppreference.com/w/cpp/types/size_t) |
+
+## Notes
+
+This concept is defined for documentation purposes but is not materialized in the library.
+
+## Example
+
+```c++
+// A size that carries an alignment hint
+template <size_t Align>
+struct aligned_size {
+    static constexpr size_t align = Align;
+    size_t size;
+    aligned_size(size_t s) : size(s) {}
+    operator size_t() const { return size; }
+};
+```
+
+[See it on Godbolt](https://godbolt.org/z/hbajKo){: .btn }
--- a/docs/extended_api/synchronization_library/make_pipeline.md
+++ b/docs/extended_api/synchronization_library/make_pipeline.md
@@ -1,0 +1,85 @@
+---
+grand_parent: Extended API
+parent: Synchronization library
+---
+
+# cuda::**make_pipeline**
+
+Defined in header [`<cuda/pipeline>`](../headers/pipeline.md)
+
+```c++
+pipeline<thread_scope_thread> make_pipeline();                                                                                       // (1)
+
+template<class Group, thread_scope Scope, uint8_t StagesCount>
+pipeline<Scope> make_pipeline(const Group & group, pipeline_shared_state<Scope, StagesCount> * shared_state);                        // (2)
+
+template<class Group, thread_scope Scope, uint8_t StagesCount>
+pipeline<Scope> make_pipeline(const Group & group, pipeline_shared_state<Scope, StagesCount> * shared_state, size_t producer_count); // (3)
+
+template<class Group, thread_scope Scope, uint8_t StagesCount>
+pipeline<Scope> make_pipeline(const Group & group, pipeline_shared_state<Scope, StagesCount> * shared_state, pipeline_role role);    // (4)
+```
+
+1. Creates a _unified pipeline_ such that the calling thread is the only participating thread and performs both producer and consumer actions.
+2. Creates a _unified pipeline_ such that all the threads in `group` are performing both producer and consumer actions.
+3. Creates a _partitioned pipeline_ such that `producer_threads` number of threads in `group` are performing producer actions while the others
+   are performing consumer actions. 
+4. Creates a _partitioned pipeline_ where each thread's role is explicitly specified.
+
+All threads in `group` acquire collective ownership of the `shared_state` storage.
+
+`make_pipeline` must be invoked by every threads in `group` such that `group::sync` may be invoked.
+
+`shared_state` and `producer_count` must be uniform across all threads in `group`, else the behavior is undefined.
+
+`producer_count` must be strictly inferior to `group::size`, else the behavior is undefined.
+
+## Template parameters
+
+| Group | a type satisfying the [_Group concept_](../concepts/group.md) |
+
+## Parameters
+
+| group          | the group of threads                                                                                                                                |
+| shared_state   | a pointer to an object of type [`cuda::pipeline_shared_state<Scope>`](./pipeline_shared_state.md) with `Scope` including all the threads in `group` |
+| producer_count | the number of _producer threads_ in the pipeline                                                                                                    |
+| role           | the role of the current thread in the pipeline                                                                                                      |
+
+## Return value
+
+A `cuda::pipeline` object.
+
+## Example
+
+```c++
+#include <cuda/pipeline>
+#include <cooperative_groups.h>
+
+// Disables `pipeline_shared_state` initialization warning
+#pragma diag_suppress static_var_with_dynamic_init
+
+__global__ void example_kernel()
+{
+    __shared__ cuda::pipeline_shared_state<cuda::thread_scope_block, 2> pss_1;
+    __shared__ cuda::pipeline_shared_state<cuda::thread_scope_block, 2> pss_2;
+    __shared__ cuda::pipeline_shared_state<cuda::thread_scope_block, 2> pss_3;
+
+    auto group = cooperative_groups::this_thread_block();
+
+    // Create a thread scoped pipeline
+    cuda::pipeline<cuda::thread_scope_thread> p_0 = cuda::make_pipeline();
+
+    // Create a unified block-scoped pipeline
+    cuda::pipeline<cuda::thread_scope_block> p_1 = cuda::make_pipeline(group, &pss_1);
+
+    // Create a partitioned block-scoped pipeline where half the threads are producers
+    size_t producer_count = group.size() / 2;
+    cuda::pipeline<cuda::thread_scope_block> p_2 = cuda::make_pipeline(group, &pss_2, producer_count);
+
+    // Create a partitioned block-scoped pipeline where all threads with an even thread_rank are producers
+    auto thread_role = (group.thread_rank() % 2) ? cuda::pipeline_role::producer : cuda::pipeline_role::consumer;
+    cuda::pipeline<cuda::thread_scope_block> p_3 = cuda::make_pipeline(group, &pss_3, thread_role);
+}
+```
+
+[See it on Godbolt](https://godbolt.org/z/Y1zv5G){: .btn }
--- a/docs/extended_api/synchronization_library/pipeline.md
+++ b/docs/extended_api/synchronization_library/pipeline.md
@@ -1,0 +1,100 @@
+---
+grand_parent: Extended API
+parent: Synchronization library
+---
+
+# cuda::**pipeline**
+
+Defined in header [`<cuda/pipeline>`](../headers/pipeline.md)
+
+```c++
+template<cuda::thread_scope Scope>
+class pipeline;
+```
+
+The class template `cuda::pipeline` provides a coordination mechanism allowing to pipeline multiple operations in a sequence of stages.
+
+A thread interacts with a _pipeline stage_ using the following pattern:
+1. Acquire the pipeline stage
+2. Commit some operations to the stage
+3. Wait for the previously committed operations to complete
+4. Release the pipeline stage
+
+For thread scopes other than `thread_scope_thread`, a [`pipeline_shared_state`](./pipeline_shared_state.md) is required to coordinate the participating threads.
+
+_Pipelines_ can be either _unified_ or _partitioned_.
+In a _unified pipeline_, all the participating threads are both producers and consumers.
+In a _partitioned pipeline_, each participating thread is either a producer or a consumer.
+
+## Template parameters
+
+### Scope
+
+A [`cuda::thread_scope`](../../api/synchronization_library/thread_scopes.md) denoting a scope including all the threads participating in the _pipeline_.
+
+## Member functions
+
+| (constructor) [deleted]                            | `pipeline` is not constructible                                                                                                                  |
+| [(destructor)](./pipeline/destructor.md)           | destroys the `pipeline`                                                                                                                          |
+| operator= [deleted]                                | `pipeline` is not assignable                                                                                                                     |
+| [producer_acquire](./pipeline/producer_acquire.md) | blocks the current thread until the next _pipeline stage_ is available                                                                           |
+| [producer_commit](./pipeline/producer_commit.md)   | commits operations previously issued by the current thread to the current _pipeline stage_                                                       |
+| [consumer_wait](./pipeline/consumer_wait.md)       | blocks the current thread until all operations committed to the current _pipeline stage_ complete                                                |
+| [consumer_wait_for](./pipeline/consumer_wait.md)   | blocks the current thread until all operations committed to the current _pipeline stage_ complete or after the specified timeout duration        |
+| [consumer_wait_until](./pipeline/consumer_wait.md) | blocks the current thread until all operations committed to the current _pipeline stage_ complete or until specified time point has been reached |
+| [consumer_release](./pipeline/consumer_release.md) | release the current _pipeline stage_                                                                                                             |
+| [quit](./pipeline/quit.md)                         | quits current thread's participation in the _pipeline_                                                                                           |
+
+## Notes
+
+A thread role cannot change during the lifetime of the pipeline object.
+
+## Example
+
+```c++
+#include <cuda/pipeline>
+#include <cooperative_groups.h>
+
+// Disables `pipeline_shared_state` initialization warning
+#pragma diag_suppress static_var_with_dynamic_init
+
+template <typename T>
+__device__ void compute(T * ptr);
+
+template <typename T>
+__global__ void example_kernel(T * global1, T * global2, size_t subset_count)
+{
+    extern __shared__ T s[];
+    auto group = cooperative_groups::this_thread_block();
+    T * shared[2] = { s, s + 2 * group.size() };
+
+    // Create a CUDA pipeline
+     constexpr unsigned stages_count = 2;
+    __shared__ cuda::pipeline_shared_state<cuda::thread_scope::thread_scope_block, stages_count> shared_state;
+    auto pipeline = cuda::make_pipeline(group, &shared_state);
+
+    // Prime the pipeline
+    pipeline.producer_acquire();
+    cuda::memcpy_async(group, shared[0], &global1[0], sizeof(T) * group.size(), pipeline);
+    cuda::memcpy_async(group, shared[0] + group.size(), &global2[0], sizeof(T) * group.size(), pipeline);
+    pipeline.producer_commit();
+
+    // Pipelined copy/compute
+    for (size_t subset = 1; subset < subset_count; ++subset) {
+        pipeline.producer_acquire();
+        cuda::memcpy_async(group, shared[subset % 2], &global1[subset * group.size()], sizeof(T) * group.size(), pipeline);
+        cuda::memcpy_async(group, shared[subset % 2] + group.size(), &global2[subset * group.size()], sizeof(T) * group.size(), pipeline);
+        pipeline.producer_commit();
+        pipeline.consumer_wait();
+        compute(shared[(subset - 1) % 2]);
+        pipeline.consumer_release();
+    }
+
+    // Drain the pipeline
+    pipeline.consumer_wait();
+    compute(shared[(subset_count - 1) % 2]);
+    pipeline.consumer_release();
+}
+```
+
+[See it on Godbolt](https://godbolt.org/z/javfx9){: .btn }
--- a/docs/extended_api/synchronization_library/pipeline/producer_acquire.md
+++ b/docs/extended_api/synchronization_library/pipeline/producer_acquire.md
@@ -1,0 +1,17 @@
+---
+nav_exclude: true
+---
+
+# cuda::pipeline\<Scope>::**producer_acquire**
+
+```c++
+void producer_acquire();
+```
+
+Blocks the current thread until the next _pipeline stage_ is available.
+
+## Notes
+
+If this method is called from a _consumer thread_ the behavior is undefined.
+
+If the pipeline is in a _quitted state_ (see [`pipeline::quit`](./quit.md)), the behavior is undefined.
--- a/docs/extended_api/synchronization_library/pipeline/quit.md
+++ b/docs/extended_api/synchronization_library/pipeline/quit.md
@@ -1,0 +1,19 @@
+---
+nav_exclude: true
+---
+
+# cuda::pipeline\<Scope>::**quit**
+
+```c++
+bool quit();
+```
+
+Quits the current thread's participation in the collective ownership of the corresponding _shared state_ ([`cuda::pipeline_shared_state`](../pipeline_shared_state.md)). Ownership of the _shared state_ is released by the last invoking thread.
+
+## Return value
+
+`true` if ownership of the _shared state_ was released, otherwise `false`.
+
+## Notes
+
+The behavior undefined if any operation other than [`~pipeline`](./destructor.md) is issued by the current thread after quitting.
--- a/docs/extended_api/synchronization_library/pipeline_consumer_wait_prior.md
+++ b/docs/extended_api/synchronization_library/pipeline_consumer_wait_prior.md
@@ -1,0 +1,57 @@
+---
+grand_parent: Extended API
+parent: Synchronization library
+---
+
+# cuda::**pipeline_consumer_wait_prior**
+
+Defined in header [`<cuda/pipeline>`](../headers/pipeline.md)
+
+```c++
+template<uint8_t Prior>
+void pipeline_consumer_wait_prior(pipeline<thread_scope_thread> & pipeline);
+```
+
+Let *Stage* be the pipeline stage `Prior` stages before the current one (counting the current one). Blocks the current
+thread until all operations committed to _pipeline stages_ up to *Stage* complete. All stages up to *Stage* (exclusive)
+are implicitly released.
+
+## Template parameters
+
+| Prior | The index of the pipeline stage *Stage* (_see above_) counting up from the current one. The index of the current stage is `0`. |
+
+## Parameters
+
+| pipeline | The thread-scoped `cuda::pipeline` object to wait on |
+
+## Notes
+
+If the pipeline is in a _quitted state_ (see [`pipeline::quit`](./pipeline/quit.md)), the behavior is undefined.
+
+## Example
+
+```c++
+#include <cuda/pipeline>
+
+__global__ void example_kernel(uint64_t * global, size_t element_count)
+{
+    extern __shared__ uint64_t shared[];
+
+    cuda::pipeline<cuda::thread_scope_thread> pipe = cuda::make_pipeline();
+    for (size_t i = 0; i < element_count; ++i) {
+        pipe.producer_acquire();
+        cuda::memcpy_async(shared + i, global + i, sizeof(*global), pipe);
+        pipe.producer_commit();
+    }
+
+    // Wait for operations committed in all stages but the last one
+    cuda::pipeline_consumer_wait_prior<1>(pipe);
+    pipe.consumer_release();
+
+    // Wait for operations committed in all stages
+    cuda::pipeline_consumer_wait_prior<0>(pipe);
+    pipe.consumer_release();
+}
+```
+
+[See it on Godbolt](https://godbolt.org/z/j83v3G){: .btn }
--- a/docs/extended_api/synchronization_library/pipeline_producer_commit.md
+++ b/docs/extended_api/synchronization_library/pipeline_producer_commit.md
@@ -1,0 +1,52 @@
+---
+grand_parent: Extended API
+parent: Synchronization library
+---
+
+# cuda::**pipeline_producer_commit**
+
+Defined in header [`<cuda/pipeline>`](../headers/pipeline.md)
+
+```c++
+template<thread_scope Scope>
+void pipeline_producer_commit(pipeline<thread_scope_thread> & pipeline, barrier<Scope> & barrier);
+```
+
+Binds operations previously issued by the current thread to the named `barrier` such that a `barrier::arrive` is performed on completion. The bind operation implicitly increments the barrier's current phase to account for the subsequent `barrier::arrive`, resulting in a net change of 0.
+
+## Parameters
+
+| pipeline | the thread-scoped `cuda::pipeline` object to wait on |
+| barrier  | the barrier to arrive on                             |
+
+## Notes
+
+If the pipeline is in a _quitted state_ (see [`pipeline::quit`](./pipeline/quit.md)), the behavior is undefined.
+
+## Example
+
+```c++
+#include <cuda/pipeline>
+
+// Disables `barrier` initialization warning
+#pragma diag_suppress static_var_with_dynamic_init
+
+__global__ void example_kernel(uint64_t * global, size_t element_count)
+{
+    extern __shared__ uint64_t shared[];
+    __shared__ cuda::barrier<cuda::thread_scope_block> barrier;
+
+    init(&barrier, 1);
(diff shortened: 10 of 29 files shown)


hcedwar (changes requested):

docs/extended_api/asynchronous_operations_library/memcpy_async.md
```diff
@@ -0,0 +1,55 @@
...
+
+```c++
+template<typename Size, thread_scope Scope>
+void memcpy_async(void * destination, void const * source, Size size, pipeline<Scope> & pipeline);                      // (3)
+
+template<typename Group, typename Size, thread_scope Scope>
+void memcpy_async(Group const & group, void * destination, void const * source, Size size, pipeline<Scope> & pipeline); // (4)
+```
+
+Asynchronously copies `size` bytes from the memory location pointed to by `source` to the memory location pointed to by `destination`.
```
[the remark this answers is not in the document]
c0riolis: I added a note to clarify this point.

docs/extended_api/asynchronous_operations_library/memcpy_async.md
```diff
@@ -0,0 +1,55 @@
...
+
+1. Binds the asynchronous copy completion to `barrier` and issues the copy in the current thread.
+2. Binds the asynchronous copy completion to `barrier` and cooperatively issues the copy across all threads in `group`.
+3. Binds the asynchronous copy completion to `pipeline` and issues the copy in the current thread
+4. Binds the asynchronous copy completion to `pipeline` and cooperatively issues the copy across all threads in `group`.
+
+## Template parameters
+
+* **Group** - a type satisfying the [_group concept_](http://LINK-TODO)
+* **Size** - [`size_t`](https://en.cppreference.com/w/c/types/size_t) or [`aligned_size_t`](http://LINK-TODO)
```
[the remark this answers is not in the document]
c0riolis: Fixed, thanks.

docs/extended_api/asynchronous_operations_library/memcpy_async.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Parameters
+
+* **group** - the group of threads 
+* **destination** - pointer to the memory location to copy to
+* **source** - pointer to the memory location to copy from
+* **size** - the number of bytes to copy
+* **barrier** - the barrier object used to wait on the copy completion
+* **pipeline** - the pipeline object used to wait on the copy completion
+
```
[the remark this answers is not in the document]
c0riolis: The `aligned_size_t` documentation clearly explains that it carries a statically defined alignment hint. I feel like it's pretty self-explanatory at this point. Maybe we can put an emphasis on it in the best practices/tuning guides?

c0riolis:
I think this is ready to be reviewed in depth. The rendered version of this commit is available here:
https://draft.dll.pe/

docs/extended_api/asynchronous_operations_library/memcpy_async.md
```diff
@@ -0,0 +1,64 @@
...
+```
+
+Asynchronously copies `size` bytes from the memory location pointed to by `source` to the memory location pointed to by `destination`.
+Both objects are reinterpreted as arrays of `unsigned char`.
+
+`cuda::memcpy_async` have similar constraints to [`std::memcpy`](https://en.cppreference.com/w/cpp/string/byte/memcpy), namely:
+* If the objects overlap, the behavior is undefined.
+* If either `destination` or `source` is an invalid or null pointer, the behavior is undefined (even if `count` is zero).
+* If the objects are potentially-overlapping or not [`TriviallyCopyable`](https://en.cppreference.com/w/cpp/named_req/TriviallyCopyable),
+  the behavior is undefined.
```
gonzalobg, line 35: If the objects are not `TriviallyCopyable` the behavior is not undefined.

docs/extended_api/asynchronous_operations_library/memcpy_async.md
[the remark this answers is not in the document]
c0riolis: I also mentioned this optimization possibility in the `aligned_size_t` example.

docs/extended_api/asynchronous_operations_library/memcpy_async.md
griwes: I think that the de jure `memcpy` is not specified to be undefined but may be so; making `memcpy_async` undefined per its spec is something I'm absolutely in favor of.

I believe that the entry points I wrote statically assert that the type is trivially copyable when possible. I believe that IFNDR would be a more modern way to spec it, but I'm unsure if the standard library uses that wording. @brycelelbach?

gnzlbg: The behavior of calling the following `well_defined` function is perfectly defined:

```c++
#include <algorithm>
#include <cstddef>
#include <cstring>
#include <type_traits>
struct NonTrivial { 
    NonTrivial() { }
};
static_assert(!std::is_trivial_v<NonTrivial>);

void well_defined() {
    NonTrivial a[2] = { NonTrivial{}, NonTrivial{} };
    std::aligned_storage_t<
      sizeof(NonTrivial[2]), 
      alignof(NonTrivial[2])
    > raw;
    std::memcpy(
        (std::byte*)&raw, 
        (std::byte*)a, 
        sizeof(NonTrivial[2])
    );
}
```

gonzalobg: The behavior of calling the following `well_defined` function is perfectly defined:

```c++
#include <algorithm>
#include <cstddef>
#include <cstring>
#include <type_traits>
struct NonTrivial { 
    NonTrivial() { }
};
static_assert(!std::is_trivial_v<NonTrivial>);

void well_defined() {
    NonTrivial a[2] = { NonTrivial{}, NonTrivial{} };
    std::aligned_storage_t<
      sizeof(NonTrivial[2]), 
      alignof(NonTrivial[2])
    > raw;
    std::memcpy(
        (std::byte*)&raw, 
        (std::byte*)a, 
        sizeof(NonTrivial[2])
    );
}
```

It just copies the bytes from `NonTrivial[2]` onto some uninitialized storage, which is fine. It never does anything that causes UB, like e.g., assuming that the objects at `raw` are live (they aren't, since we haven't called their constructors). 

A lot of code uses `memcpy` to copy raw bytes around like this (including bytes of non-trivial objects, uninitialized padding bytes, etc.). None of this is UB per se, and it is not UB for `memcpy`.

What's the rationale for making this "library UB" for `memcpy_async`?

griwes: Trivial and trivially copyable are different properties. The type above should be trivially copyable, no?

griwes: Non trivially copyable types are not guaranteed to be copyable with memcpy anywhere in the standard. See [basic.types.general] calling out trivially copyable types wherever it mentions memcpy.

gonzalobg: >  The type above should be trivially copyable, no?

It is. Adding a copy constructor makes it non-trivially copyable. I've fixed this above.

griwes: Yes, and that makes it illegible for the additional allowances made by [basic.types.general], which is why it's UB here.

Non trivially copyable types are fragile when you copy them in a way different than by, well, copying them. I just noticed your code sample uses the power of std::byte to alias things, and it's invoking memcpy with a byte pointer, not a `NonTrivial` pointer, which means that it is copying the representation, not the object itself; that is fine and not prohibited by the call being UB if the objects are not trivially copyable. I admit the allowed cases of aliasing (through byte and chars) make this really muddy.

I tried to come up with a wording that makes this clearer, but I'm failing so far since the signatures don't talk about the actual pointer types, just about void pointers... But this follows what cppreference says about memcpy modulo turning "unspecified but maybe undefined" into "undefined", which IMHO is a much better place to be.

gonzalobg: > Non trivially copyable types are not guaranteed to be copyable with memcpy anywhere in the standard.

I disagree.

The standard guarantees that all objects have a representation, which consist of the sequence of sizeof(T) `unsigned char` objects taken up by the object (see [basic.types.general.4.1](http://eel.is/c++draft/basic.types.general#4.sentence-1)).

For `memcpy`, AFAICT the standard defers to the C standard library: [[cstring.syn.1]](http://eel.is/c++draft/cstring.syn#1), which just specifies that it copies N bytes, which are there, and has no notion of whether the types are trivially copyable or not, or C++ objects / lifetimes.

The `memcpy` above arguably creates new live objects of type `std::byte` in the `raw` storage; that's ok AFAICT. 

In [basic.types.general], the standard guarantees that for trivially copyable types, a round-trip `memcpy` to an array of `char`/`byte`/etc. preserves the object's _value_. It does not guarantee this anywhere for types that are not trivially copyable, but the example above does not use that.

gonzalobg: > I just noticed your code sample uses the power of std::byte to alias things, and it's invoking memcpy with a byte pointer, not a NonTrivial pointer

The signature of memcpy is `memcpy(void*, void*, size_t)`, not `memcpy(T*, T*, size_t)` (`memcpy` is not `std::copy`), so I'm not sure this matters. Whatever pointer type you take, there is a type conversion to `void*` that's going to happen anyways.

griwes: It does matter; pointer values are the addresses in memory imbued with what object they point into.

gonzalobg: So? `memcpy` is just using that address to copy a representation.

gonzalobg: > I just noticed your code sample uses the power of std::byte to alias things,

Removing `std::byte` from the example still does not exhibit UB AFAICT:

```c++
#include <algorithm>
#include <cstddef>
#include <cstring>
#include <type_traits>
struct NonTrivial { NonTrivial() { } NonTrivial(NonTrivial const&) { }};
static_assert(!std::is_trivially_copyable_v<NonTrivial>);


void well_defined() {
    NonTrivial a[2] = { NonTrivial{}, NonTrivial{} };
    std::aligned_storage_t<
      sizeof(NonTrivial[2]), 
      alignof(NonTrivial[2])
    > raw;
    std::memcpy((void*)&raw, (void*)a, sizeof(NonTrivial[2]);
    );
}
```

c0riolis: Taking a more pragmatic approach for a second, why would you **ever** want to do that (extracting a non-trivially-copyable object representation)? You cannot re-materialize a valid object from it.

Also, [[intro.object.8.4]](http://eel.is/c++draft/basic#intro.object-8.4) seems to indicate that:

> An object of trivially copyable or standard-layout type ([basic.types]) shall occupy contiguous bytes of storage.

I see no such requirement for non-trivially-copyable types, which seems incompatible with your claim.

griwes: @gonzalobg I am not convinced that is well-defined; can you provide me with quotes of the normative text that define the behavior?

gonzalobg: > You cannot re-materialize a valid object from it.

Whether you can depends on the object (simple example: an object with a constructor that takes its raw representation can rematerialize itself from it).

> I see no such requirement for non-trivially-copyable types, which seems incompatible with your claim.

AFAICT that just means these objects can have padding, and `memcpy` can copy padding bytes just fine. There is no other option, since as mentioned above, [basic.types.general.4.1](http://eel.is/c++draft/basic.types.general#4.sentence-1) guarantees that all objects occupy a _sequence_ of `sizeof(T)` `unsigned char`s.

> I am not convinced that is well-defined; can you provide me with quotes of the normative text that define the behavior?

* [basic.types.general.4.1](http://eel.is/c++draft/basic.types.general#4.sentence-1) guarantees that all objects occupy a _sequence_ of `sizeof(T)` `unsigned char`s.
* [[cstring.syn.1]](http://eel.is/c++draft/cstring.syn#1 defers to [7.24.2.1](https://port70.net/~nsz/c/c11/n1570.html#7.24.2.1), which specifies `std::memcpy`, which just "copies n _characters_" from one object to another (in this case, from a `NonTrivial[2]` to a `unsigned char[sizeof(NonTrivial[2])];`. 

I don't think there is anything else to define: memcpy copies characters of the representation, objects have representations, hence you can copy from one representation to another. 

C++ adds lifetimes, so when you copy into a new representation, the old one whose storage is being reused is dead (which is fine for `unsigned char`, it does not have destructors), and a new object is created (which is also fine for `unsigned char`, since it is trivially copyable and constructible).

gonzalobg: I think this example can be further reduced to whether: 

```c++
NonTrivial foo{};
unsigned char c = *((unsigned char*)&foo);
```
is UB or not since this is essentially what memcpy semantically does (it then writes `c` somewhere else; nothing here technically prevents preserving provenance).

gonzalobg: See http://wg21.link/p18392r2 , which hints that there is unanimous consensus that accessing the entire object representation through a char-like type for contiguous objects should be well-defined and fixed as DR (defect report?).

It forwards to http://wg21.link/p1945 which address non-contiguous types (virtual base sub-objects), but in this particular case, the type in the example is contiguous, or the example can be easily fixed to only run if it is contiguous.

griwes: I don't think that's the same issue, but I'm starting to lean towards your reading of the words for std::memcpy, even though the 9899 definition is somewhat problematic given how 14882 defines objects.

Having conceded that though, I still believe that making such calls should be UB for memcpy_async. There's little you can do with the resulting memory location; if the type is standard layout, you could extract a subobject's offset and read its value if it's trivially copyable, but then you can just copy the subobject instead. (Non trivially copyable objects don't have values per the standardese definition of the word.)

Is there any particular use cases that are stopped by not making the calls with non trivially copyable types valid for memcpy_async? I'd rather we build in some safety into the C-like APIs we provide unless there's some compelling reason to do so.

gonzalobg: > I'd rather we build in some safety into the C-like APIs we provide unless there's some compelling reason to do so.

Making this "library UB" or not is a design trade-off. Is the value proposition for making this library UB worth it? AFAICT, the costs of this trade-offs are:

- how do we teach it: yet another case of UB, need to mention it when mentioning `memcpy_async` instead of just saying "exact same semantics as `memcpy`
- can we catch it and how: I don't think we can catch it at the library level (implementing this as `memcpy_async<T>(T*,T*,size_t)` and failing on T=NonTriviallyCopyable has a bunch of issues), catching this at the compiler level would be hardish and not worth the effort, catching this at the sanitizer level would be very hard, so I think this puts this in the bag of UB that we can't diagnose.
- complicates our docs: it would be simpler to just say "exact same semantics as `memcpy`", whatever those might be, and if a DR lands that clarifies the behavior that all existing compilers implement, that automatically would apply.

and the value of making this UB is: 

- optimizations: if we make this an intrinsic (which we probably should if it isn't already), are there any optimizations that this allows beyond the tautological  "dead code removal because this would be UB" ?

What other value those it add? 

---

I think the simplest, future-proof thing to do here is to just say that this has the exact same semantics as `memcpy`. It doesn't help users much, because the semantics of `memcpy` aren't 100% clear in the standard, but that's the standard fault, and it is easier to solve that via a defect report or a paper, than try to pick one or the other side here.

griwes: Well, memcpy_async is supposed to be a replacement for assigning from global memory to shared memory; if we allow byte-wise copies to be done for non trivially copyable types, we'll enable a class of bugs where users blindly do the recommended thing and get invalid behavior. By diagnosing the cases that are suspect (as the implementation does right now), we eliminate a subset of that class of bugs.

@hcedwar can you weigh in here from the point of view of platform exposure?

griwes: We don't want to fall back to magically having our semantics changed from under us when a DR is applied to the standard. I really want to be able to vet any DR'd changes against the intent of our platform exposure for the parts of our stack outside of `cuda::std::`.

Remember that unlike memcpy, memcpy_async isn't merely a tool to punch holes in the object model; it's a hardware feature exposure mechanism. It needs to primarily serve as such, instead of being forced into the shape of a standard tool that is not serving such a goal.

gonzalobg: > We don't want to fall back to magically having our semantics changed from under us when a DR is applied to the standard. 

I would actually want that. The alternative is to live with very subtle differences in semantics between C++ and CUDA C++. That would add a significant teaching, documentation, and specification cost, as well as confusion, which would lead to bugs. 

> By diagnosing the cases that are suspect (as the implementation does right now),

As mentioned, this introduces a bunch of issues. Some examples are "violating our own spec for the signature of `memcpy_async`" (which isn't implemented as we specify it anymore), "breaking SFINAE for `memcpy_async`" (e.g. [this example](https://gcc.godbolt.org/z/cGs8hW) never copies a non-trivially copyable type with `memcpy_async`, and according to our spec it should compile, but it doesn't, because our implementation is non-conforming). There are probably more consequences of doing this.

> we'll enable a class of bugs where users blindly do the recommended thing and 
get invalid behavior. 
>
>  Remember that unlike memcpy, memcpy_async isn't merely a tool to punch holes in the object model; it's a hardware feature exposure mechanism. 

The C++ standard delegates the specification of `memcpy` to the C standard, in which it is essentially a tool for copying raw bytes, which is quite close to what the hardware does (the hardware doesn't care much about the bytes being trivially-copyable). 

It appears that what you want is to go beyond that, and have a safer API. In C++, that API is `std::copy`. An async version of `std::copy` that does the right thing depending on TriviallyCopyable, ContiguousRange, etc. can probably be implemented on top of a raw `memcpy_async` API.  

The only reason my original example would be "buggy" in CUDA C++ using `memcpy_async` is because we say that it exhibits "library UB". In C++, that non-trivially copyable type has contiguous storage on all implementations, and a paranoid user could write code to test that, which means that the code would not be buggy. Removing "library UB" from the `async_memcpy` specification would remove those bugs. 

I agree that a safer asynchronous copy API makes sense (e.g. `std::copy_async`). I don't know whether the value proposition of trying to bolt a degree of safety onto a `memcpy`-like API does (it looks to me that we are trading off catching some bugs by introducing others).

hcedwar:
Given that _memcpy_async is supposed to be a replacement for assigning from global memory to shared memory_ ( https://github.com/NVIDIA/libcudacxx/pull/53#discussion_r520364317 ) the statement
_The implementation may perform the copy synchronously in some implementation defined cases._
is inadequate for users to understand when asynchronous functionality is or is not possible.
Emphasis should be on **how** GPU customer-developers can obtain value (asynchronous functionality), not on how to introduce minimal wording change from `std::memcpy`.

griwes:
@hcedwar thanks, that's good feedback, but can you contribute to the long comment thread above? Thanks!

docs/extended_api/asynchronous_operations_library/memcpy_async.md
griwes: > I would actually want that. The alternative is to live with very subtle differences in semantics between C++ and CUDA C++. That would add a significant teaching, documentation, and specification cost, as well as confusion, which would lead to bugs.

That's a fine stance to have, and you are free to have that stance, but that is not the stance _we_ have for functionalities that are not direct extensions of C++ APIs.

> As mentioned, this introduces a bunch of issues. Some examples are "violating our own spec for the signature of memcpy_async" (which isn't implemented as we specify it anymore),

I don't understand what this means. How do we break the spec? Compile-time UB is still a thing, failing to compile *is* a valid way to handle UB; see [[defns.undefined]](https://eel.is/c++draft/defns.undefined). Please elaborate on how you think we break the spec.

> "breaking SFINAE for memcpy_async" (e.g. this example never copies a non-trivially copyable type with memcpy_async, and according to our spec it should compile, but it doesn't, because our implementation is non-conforming).

Where does our spec say it should compile? The spec presents overloads that accept `void *` arguments and nowhere promises any templated overloads that you can SFINAE on. The example you've linked is not valid, because it tries to SFINAE on a function call that is not specified to give you SFINAE. What part of the docs made you think the example is valid? It's a problem with that part of the spec, not with this.

> The C++ standard delegates the specification of memcpy to the C standard, in which it is essentially a tool for copying raw bytes, which is quite close to what the hardware does (the hardware doesn't care much about the bytes being trivially-copyable).

That's fine, and if/when we expose `cuda::std::memcpy`, we will follow that specification. `memcpy_async` is a separate feature with its own design that's _based_ on `std::memcpy`, and therefore has its own specification.

> The only reason my original example would be "buggy" in CUDA C++ using memcpy_async is because we say that it exhibits "library UB". In C++, that non-trivially copyable type has contiguous storage on all implementations, and a paranoid user could write code to test that, which means that the code would not be buggy. Removing "library UB" from the async_memcpy specification would remove those bugs.

Your earlier example is not driving this decision; it's the original use case for hardware exposure that is driving this decision. Consider:

```cuda
template<typename T>
__global__ void kernel(T * global_mem, std::size_t num_elements) {
    __shared__ T shared_mem[THREAD_COUNT];
    for (auto i = 0; i < num_elements / THREAD_COUNT; ++i) {
        shared_mem[threadIdx.x] = global_mem[blockDim.x * i + threadIdx.x];
        // ... do computations
    }
    // ... handle the tail of the data set
}
```

`memcpy_async` is meant _primarily_ to enable the user to rewrite functions like this. Most functions like this will probably just expect to work on integral or floating-point types, which are trivially copyable. But suppose that it's meant to work with a multitude of types, and the user rewrites (as our messanging says they should!) the `shared_mem[threadIdx.x] = global_mem[blockDim.X * i + threadIdx.x];` line with a call to memcpy_async (probably in the vein of `cuda::memcpy_async(cg::this_thread_block(), shared_mem, global_mem[blockIdx.x * i], sizeof(T), pipeline);`. This is _good_, but if we allowed non-trivially-copyable types, this may introduce a _very subtle_ bug of doing a bytewise copy of a type with custom copy semantics, and the rewrite has now changed the semantics of the program in question... ...despite it being a rewrite that we recommend users do. Whoopsie. Not a great look. If this fails to compile with a very explicit error message (except on GCC 4.8, but that can't be helped), we've saved the user from rewriting a kernel into a form that breaks its semantics.

If, however, the user wants to still force it... they can just cast to `void *` or `std::byte *` (and wrap the size in `cuda::aligned_size_t`).

If we ever arrive in a place where there's more _actual_ use-cases for `memcpy_async` than the above rewrite, I'll happily come back to this discussion and possibly revisit it. After all, removing a requirement like this from a function is a backwards compatible change, so when circumstances of `memcpy_async` change, we can very easily go in this direction if it is deemed reasonable.

docs/extended_api/synchronization_library/pipeline.md
```diff
@@ -0,0 +1,104 @@
...
+                       &global2[0], sizeof(T) * group.size(), pipeline);
+    pipeline.producer_commit();
+
+    // Pipelined copy/compute
+    for (size_t subset = 1; subset < subset_count; ++subset) {
+        pipeline.producer_acquire();
+        cuda::memcpy_async(group, shared[subset % 2],
+                           &global1[subset * group.size()], sizeof(T) * group.size(), pipeline);
+        cuda::memcpy_async(group, shared[subset % 2] + group.size(),
+                           &global2[subset * group.size()], sizeof(T) * group.size(), pipeline);
```
griwes, line 90: Reformat the calls to `memcpy_async` so that a line split does not occur between the destination and source arguments.

docs/extended_api/synchronization_library/pipeline/quit.md
```diff
@@ -0,0 +1,19 @@
...
+
+Quits the current thread's participation in the collective ownership of the corresponding _shared state_ ([`cuda::pipeline_shared_state`](../pipeline_shared_state.md)). Ownership of the _shared state_ is released by the last invoking thread.
+
+## Return value
+
+`true` if ownership of the _shared state_ was released, otherwise `false`.
+
+## Notes
+
+The behavior undefined if any operation other than [`~pipeline`](./destructor.md) is issued by the current thread after quitting.
```
griwes, line 19: Can we also add this note as a precondition to all the functions that are no longer valid to call after `quit`?

docs/extended_api/synchronization_library/pipeline_consumer_wait_prior.md
```diff
@@ -0,0 +1,52 @@
...
+
+Defined in header [`<cuda/pipeline>`](../headers/pipeline.md)
+
+```c++
+template<uint8_t Prior>
+void pipeline_consumer_wait_prior(pipeline<thread_scope_thread> & pipeline);
+```
+
+Blocks the current thread until all operations committed to _pipelines stages_ sequenced before the `Prior` last one complete. All stages up to `Prior` (excluded)
+are implicitly released.
```
griwes, line 16: I get what this says, but probably only because I know what it means to say ;) How about:

"Let *Stage* be the pipeline stage `Prior` stages before the current one (counting the current one). Blocks the current thread until add operations committed to _pipeline stages_ up to *Stage* complete. All stages up to *Stage* (inclusive) are implicitly released." (Double check for off-by-ones in this wording.)

I'm trying to introduce an actual _stage_ that we can talk about, instead of talking about "`Prior` last one", which is a really confusing formulation when you don't already know what it means.

docs/extended_api/synchronization_library/pipeline_consumer_wait_prior.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Template parameters
+
+| Prior | The Nth latest stage to wait for |
```
griwes, line 20:
```suggestion
| Prior | The index of the pipeline stage *Stage* (_see above_) counting up from the current one |
```

I admit that this is a bit wonky, but I think it's an improvement - it doesn't introduce `N` with no clear definition of what it is. We may need to also say "the index of the current stage is 0" or something here.

(TBQH I really hate this functionality, only because of how hard it is to word _what it does_ in English.)

docs/extended_api/synchronization_library/pipeline_producer_commit.md
```diff
@@ -0,0 +1,47 @@
...
+```c++
+template<thread_scope Scope>
+void pipeline_producer_commit(pipeline<thread_scope_thread> & pipeline, barrier<Scope> & barrier);
+```
+
+Binds operations previously issued by the current thread to the named `barrier` such that a `barrier::arrive` is performed on completion. The bind operation implicitly increments the barrier's current phase to account for the subsequent `barrier::arrive`, resulting in a net change of 0.
+
+## Parameters
+
+| pipeline | the thread-scoped `cuda::pipeline` object to wait on |
```
griwes, line 19: Missing `barrier`.

docs/extended_api/synchronization_library/pipeline_role.md
```diff
@@ -0,0 +1,25 @@
...
+grand_parent: Extended API
+parent: Synchronization library
+---
+
+# cuda::**pipeline_role**
+
+Defined in header [`<cuda/pipeline>`](../headers/pipeline.md)
+
+```c++
+enum pipeline_role : /* unspecified */ {
```
griwes, line 11:
```suggestion
enum class pipeline_role : /* unspecified */ {
```

griwes (changes requested):

docs/extended_api/asynchronous_operations_library/memcpy_async.md
```diff
@@ -0,0 +1,64 @@
...
+| pipeline    | the pipeline object used to wait on the copy completion |
+
+## Notes
+
+The implementation may perform the copy synchronously in some implementation defined cases.
+
+## Example
+
+```c++
+TODO
```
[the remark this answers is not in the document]
c0riolis: This is on purpose, see [this comment](https://github.com/NVIDIA/libcudacxx/pull/53#issuecomment-723025650).

docs/extended_api/concepts/group.md
```diff
@@ -0,0 +1,46 @@
+---
+grand_parent: Extended API
+parent: Concepts
+---
+
+# Group
+
+```c++
+struct Group {
+  static constexpr cuda::thread_scope thread_scope = /*implementation-defined*/;
```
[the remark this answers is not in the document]
c0riolis: > I'm also unsure if this way of spelling it doesn't give the impression that the implementation defines what all Groups must define as the value here. Unsure what's a better way to spell it though.

I agree it might be confusing. We could just remove the `implementation-defined` comment and write it like so:
> `static constexpr cuda::thread_scope thread_scope;`

Thoughts?

docs/extended_api/concepts/group.md
```diff
@@ -0,0 +1,46 @@
...
+  integral thread_rank() const;
+  void sync() const;
+};
+```
+
+The _Group concept_ defines the requirements of a type that represents a group of cooperating threads.
+
+## Data members
+
+| thread_scope | the scope at which `Group::sync()` synchronizes memory operations |
```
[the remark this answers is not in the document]
c0riolis: Good point, fixed.

docs/extended_api/concepts/group.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Example
+
+```c++
+#include <cuda/atomic>
+
+struct single_thread_group {
+    static constexpr cuda::thread_scope thread_scope = cuda::thread_scope::thread_scope_thread;
+    size_t size() const { return 1; }
+    size_t thread_rank() const { return 1; }
```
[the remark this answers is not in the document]
c0riolis: Nice catch, fixed.

docs/extended_api/concepts/shape.md
```diff
@@ -0,0 +1,37 @@
+---
+grand_parent: Extended API
+parent: Concepts
+---
+
+# Shape
+
+```c++
+struct Shape {
+    operator size_t const;
```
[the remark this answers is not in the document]
c0riolis: Fixed, thanks.

docs/extended_api/synchronization_library/make_pipeline.md
```diff
@@ -0,0 +1,85 @@
...
+## Parameters
+
+| group          | the group of threads                                                                                                                   |
+| shared_state   | an object of type [`cuda::pipeline_shared_state<Scope>`](./pipeline_shared_state.md) with `Scope` including all the threads in `group` |
+| producer_count | the number of _producer threads_ in the pipeline                                                                                       |
+| role           | the role of the current thread in the pipeline                                                                                         |
+
+## Return value
+
+A thread-local `cuda::pipeline` object.
```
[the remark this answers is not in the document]
c0riolis: I removed `thread-local`.

docs/extended_api/synchronization_library/pipeline.md
```diff
@@ -0,0 +1,104 @@
...
+1. Acquire the pipeline stage
+2. Commit some operations to the stage
+3. Wait for the previously committed operations to complete
+4. Release the pipeline stage
+
+For thread scopes other than `thread_scope_thread`, a [`pipeline_shared_state`](./pipeline_shared_state.md) is required to coordinate the participating threads.
+
+_Pipelines_ can be either _unified_ or _partitioned_.
+In a _unified pipeline_, all the participating threads are both producers and consumers.
+In a _partitioned pipeline_, each participating thread is either a producer or (exclusive) a consumer.
```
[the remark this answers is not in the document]
c0riolis: Good point.

docs/extended_api/synchronization_library/pipeline_consumer_wait_prior.md
c0riolis: Agreed, your version is better, I made two minor changes:
* until ~add~ **all** operations committed
* All stages up to Stage (~inclusive~ **exclusive**) are implicitly released.

docs/extended_api/synchronization_library/pipeline/quit.md
c0riolis: I avoided doing that because it would extend to operations on the pipeline (e.g. `memcpy_async`) and I judged it too verbose. Are you thinking about only doing it for `pipeline` member functions?

c0riolis:
Thanks for the review @griwes, I addressed most of your comments and replied to a couple of others for clarification.

c0riolis:
> Given that _memcpy_async is supposed to be a replacement for assigning from global memory to shared memory_ ( [#53 (comment)](https://github.com/NVIDIA/libcudacxx/pull/53#discussion_r520364317) ) the statement
> _The implementation may perform the copy synchronously in some implementation defined cases._
> is inadequate for users to understand when asynchronous functionality is or is not possible.
> Emphasis should be on **how** GPU customer-developers can obtain value (asynchronous functionality), not on how to introduce minimal wording change from `std::memcpy`.

@hcedwar: I do not think that this is the role of this document to specify in which cases the implementation actually exhibits asynchronous behavior, especially since it varies depending on the architecture.

In my opinion, this information belongs to one of our other guides (programming guide, best practices guide, architecture-specific tuning guide).

docs/extended_api/asynchronous_operations_library/memcpy_async.md
gonzalobg: > Where does our spec say it should compile? The spec presents overloads that accept void * arguments and nowhere promises any templated overloads that you can SFINAE on. 

The spec only provides `void*` overloads. If the implementation would match the spec, my _valid_ SFINAE example that never invokes UB even according to our restrictive spec would compile just fine. The reason it doesn't is because we provide more overloads than what the spec provides. 

> If, however, the user wants to still force it... they can just cast to void * or std::byte * (and wrap the size in cuda::aligned_size_t).

[My example](https://gcc.godbolt.org/z/cGs8hW) is well-defined according to our spec (it never calls `memcpy_async` with a non-trivially copyable type), yet we reject it. Having to use `void*` here is IMO a bug in the implementation.

Other examples that call `memcpy_async` with a non-trivially copyable type are UB according to our spec. We shouldn't be recommending to use `void*` to workaround that.

---

Adding "safety" to `memcpy` is a trade-off. I think that the costs of doing this are (at least):

- deviating from standard C++ `memcpy` semantics, 
- adding "library UB" that's impossible to check for properly (we can't check `void*`), 
- deviating from our documented API, and
- rejecting a subset of valid programs.

I don't think these costs are worth the value of catching the subset of bugs mentioned. This is a value proposition, if you and others think that these costs are worth their value, I respect that. 

C++ solution to this problem is `std::copy`, which is implemented on top of the "dangerous" `std::memcpy`. I think it would be better to have a safe `copy_async` API that's actually safe to use, instead of paying a high cost to get `memcpy_async` only half-way there.

docs/extended_api/synchronization_library/pipeline/quit.md
griwes: We should do it everywhere where we take a pipeline and have a precondition on the pipeline. We should avoid burrying it in notes of just a single member function; that's easy to overlook.

docs/extended_api/concepts/group.md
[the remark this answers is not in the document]
griwes: Yeah, I don't hate that.

docs/extended_api/asynchronous_operations_library/memcpy_async.md
griwes: > The spec only provides void* overloads. If the implementation would match the spec, my valid SFINAE example i that never invokes UB even according to our restrictive spec would compile and execute just fine. The reason it doesn't is because we provide more overloads than what the spec provides.

Regarding "more overloads": since `memcpy_async` is not an [_addressable function_](http://eel.is/c++draft/library#namespace.std-6), you are not allowed to take its address, which makes it impossible for you to tell what overloads are present in the implementation and whether they exactly match the overloads in the API surface or not. The spec defines the API; the implementation implements that API (the total sum of the overloads present in the implementation is the full API surface, and there's no ambiguitites between them in valid calls to the API surface).

> My example is well-defined according to our spec (it never calls memcpy_async with a non-trivially copyable type), yet we reject it.

It's very interesting that you think it does not call memcpy_async with a non trivially copyable type, because the diagnostic you are getting comes from a line in the program that invokes memcpy_async with a non trivially copyable type. The call on line 39 is unconditional; the "trait" you defined is unconditionally true; that means that line 39 unconditionally leads to executing line 22, which is undefined behavior per the spec. Unless of course you want to argue that this is only true if `foo` is ever invoked, but that's playing semantics for no useful reason.


> deviating from standard C++ memcpy semantics, etc.

We necessarily deviate from memcpy semantics.

> adding "library UB" that's impossible to check for properly (we can't check void*),

It is possible to check for the trap in the intended and suggested use cases.

> deviating from our documented API,

Citation still needed. Repeating the assertion that your example is well-defined according to the spec does not make it so.

> rejecting a subset of valid programs. and

Ditto.

> higher teachability, documentation, and implementation costs due to all of the above.

This is highly subjective and I don't think I agree with any of these.


--------------------------


I appreciate the feedback, and I also appreciate the course this discussion has run; I think I am now closer to agreeing with you about how well defined `std::memcpy` is for a certain class of calls (though I'm pretty sure I'm not there yet, at least not fully ;)). If/when the intented use cases of `memcpy_async` change, we will revisit the conscious decisions made about the requirements placed on calls to `memcpy_async`. For now, since I feel like we have reached a point in the discussion where we are bound to run in circles, I'll resolve this conversation, in the interests of keeping the PR view clearer than it has become thanks to the volume of this thread.

gonzalobg: > It's very interesting that you think it does not call memcpy_async with a non trivially copyable type, because the diagnostic you are getting comes from a line in the program that invokes memcpy_async with a non trivially copyable type. 

I understand that this is hard to see in that example. [This program shows the crux of the issue (godbolt)](https://gcc.godbolt.org/z/W47Yxx)):  

```c++
template <typename T>
__global__ void kernel(...) {
    if (std::is_trivially_copyable_v<T>) {
        cuda::memcpy_async(g, shared, global, sizeof(NTC), b);
    } else {
        shared[i] = global[i];
    } 
}
```

I believe this program is valid according to our specification and should be accepted. Yet our current implementation rejects it. 

This program never calls `memcpy_async` with a non-trivially copyable type and does not have UB. Yet the compilation error rejects it under that premise. 

I hope this example makes the issue clear.

docs/extended_api/asynchronous_operations_library/aligned_size_t.md
```diff
@@ -0,0 +1,61 @@
...
+## Member functions
+
+| [(constructor)](./aligned_size_t/constructor.md) | constructs an _aligned size_                                                      |
+| (destructor) [implicitly declared]               | trivial implicit destructor                                                       |
+| operator= [implicitly declared]                  | trivial implicit copy/move assignment                                             |
+| operator size_t                                  | implicit conversion to [`size_t`](https://en.cppreference.com/w/cpp/types/size_t) |
+
+## Notes
+
+If `value` is not a multiple of `align` the behavior is undefined.
```
gonzalobg, line 37: Is `align=16` and `value=0` ok ? (e.g. does this support sizes that are smaller than the alignment)

Maybe:

```suggestion
If `value` is smaller than `align` or `value` is not a multiple of `align`, the behavior is undefined.
```

docs/extended_api/asynchronous_operations_library/aligned_size_t.md
```diff
@@ -0,0 +1,61 @@
...
+
+If `Alignment` is not a [valid alignment](https://en.cppreference.com/w/c/language/object#Alignment) the behavior is undefined.
+
```
gonzalobg, line 40: How does this interact with the "validity" requirement of `memcpy_async` ?

For example, here:

```c++
T a, b;
cuda::memcpy_async(g, &b, &a, cuda::aligned_size_t<alignof(T) * 2>(0), b);
assert(&a %  (alignof(T) * 2) == 0);
```

the application is stating that the alignment of the `&a` and `&b` pointers is `2 * alignof(T)`, which might or might not be the case depending on what addresses these variables have at run-time (if any). Is this stating that such an alignment requirement is required for those addresses to be valid? 

That is, would it be a sound program transformation to remove the `assert` that (hopefully) is checking that this is indeed the case since we are stating that it is?

docs/extended_api/asynchronous_operations_library/memcpy_async.md
griwes: That's a better example, thanks; but getting rid of a diagnostic in that case is also simple: use if constexpr instead.

I've heard the solutions you've been proposing for several comments. We will relitigate if/when the intended use cases for memcpy_async change.

docs/extended_api/asynchronous_operations_library/aligned_size_t.md
griwes: No, memcpy_async is explicitly UB when the size is 0. The wording here is fine, since it does allow zero as is (and the suggested change would allow, say, 7, for alignment of 8, which is not intended).

docs/extended_api/asynchronous_operations_library/aligned_size_t.md
griwes: Correct. If the user uses aligned_size_t, they are providing an assumption that the pointers (and the size) are aligned at least to that alignment, and if that's not the case, it's on the user (and UB).

gonzalobg: Thanks. Should we call this out a bit more explicitly somewhere ?

docs/extended_api/asynchronous_operations_library/aligned_size_t.md
gonzalobg: > No, memcpy_async is explicitly UB when the size is 0. 

Where is this stated? 

In the docs of `memcpy_async` below it says:

>  If either `destination` or `source` is an invalid or null pointer, the behavior is undefined (even if `count` is zero).

That suggests that a `count == 0` is ok as long as `destination` and `source` are valid.

docs/extended_api/asynchronous_operations_library/memcpy_async.md
gonzalobg: > That's a better example, thanks; but getting rid of a diagnostic in that case is also simple: use if constexpr instead.

The issue remains that a valid program is rejected. If this program is not valid, the docs should IMO say so.

Also, replacing SFINAE with `if constexpr` isn't trivial, so I don't think that's a good solution.

docs/extended_api/asynchronous_operations_library/aligned_size_t.md
griwes: Whoops, I got carried away. That'll teach me to not respond when it's way too late for me to be awake anyway :P 

Anyway, the second part stands.

docs/extended_api/asynchronous_operations_library/aligned_size_t.md
griwes: Yeah, we should. Good catch. I thought we did.

cc @c0riolis

docs/extended_api/asynchronous_operations_library/memcpy_async.md
griwes: Okay, that is a fair point. I wish you could've started with examples like that :)

@c0riolis can you split the second part of the last paragraph off into its own remark, and reword it to something like, "a program that contains a call to `memcpy_async` with arguments pointing to objects of types that are not TriviallyCopyable is ill-formed, no diagnostic required."?

docs/extended_api/asynchronous_operations_library/aligned_size_t.md
gonzalobg: @c0riolis this issue can be marked as resolved.

docs/extended_api/asynchronous_operations_library/memcpy_async.md
gonzalobg: > I wish you could've started with examples like that :)

Apologies, really. There are a couple of issues mixed in this thread, and github is not great at keeping different threads of discussion separate (or cleaning them up afterwards). 

I didn't know that it was intentional to deviate from `std::memcpy` here, so we had to resolve that, and then it turned out that I disagree with that design. For me that's not an issue, but raising the tradeoffs I see and understanding the ones that others have raised is. In retrospect, a github thread is not the best avenue for that.

griwes: Yeah, that sentiment is a part of what I meant. I didn't intend that as an accusation of any sort ;)

docs/extended_api/asynchronous_operations_library/memcpy_async.md
```diff
@@ -0,0 +1,63 @@
...
+| group       | the group of threads                                    |
+| destination | pointer to the memory location to copy to               |
+| source      | pointer to the memory location to copy from             |
+| size        | the number of bytes to copy                             |
+| barrier     | the barrier object used to wait on the copy completion  |
+| pipeline    | the pipeline object used to wait on the copy completion |
+
+## Notes
+
+The implementation may perform the copy synchronously in some implementation defined cases.
```
gonzalobg, line 57: If this is _implementation defined_, then we have to define what these cases are, because we are the implementation.

c0riolis: I just removed the note, we can address this topic in the performance tuning guide.

docs/extended_api/asynchronous_operations_library/aligned_size_t.md
c0riolis: Done, thanks.

docs/extended_api/synchronization_library/pipeline/quit.md
c0riolis: Done.

docs/extended_api/asynchronous_operations_library/memcpy_async.md
c0riolis: > @c0riolis can you split the second part of the last paragraph off into its own remark, and reword it to something like, "a program that contains a call to memcpy_async with arguments pointing to objects of types that are not TriviallyCopyable is ill-formed, no diagnostic required."?
Done.

docs/extended_api/synchronization_library/pipeline/producer_acquire.md
```diff
@@ -0,0 +1,17 @@
...
+nav_exclude: true
+---
+
+# cuda::pipeline\<Scope>::**producer_acquire**
+
+```c++
+void producer_acquire();
+```
+
+Blocks the current thread until the next _pipeline stage_ is available.
```
gnzlbg, line 11: Do we document the relationship between `producer_acquire` and `cuda::pipeline_shared_state`'s `StagesCount` template parameter somewhere?

If not, I think we should document that somewhere.

gonzalobg, line 11: Do we document the relationship between `producer_acquire` and `cuda::pipeline_shared_state`'s `StagesCount` template parameter somewhere?

If not, I think we should.

gonzalobg: For example, we guarantee that here:

```c++
__global__ void foo() {
constexpr unsigned stages_count = 2;
__shared__ cuda::pipeline_shared_state<cuda::thread_scope::thread_scope_block, stages_count> shared_state;
￼auto pipeline = cuda::make_pipeline(group, &shared_state);
pipeline.producer_acquire();
...
pipeline.producer_commit();
pipeline.producer_acquire(); // This acquire never blocks
...
```

the second call to `producer_acquire` never blocks because the `pipeline` has a "capacity` for holding two Stages.

If we change the `StageCount` to `1`: 

```c++
__global__ void foo() {
constexpr unsigned stages_count = 1;
__shared__ cuda::pipeline_shared_state<cuda::thread_scope::thread_scope_block, stages_count> shared_state;
￼auto pipeline = cuda::make_pipeline(group, &shared_state);
pipeline.producer_acquire();
...
pipeline.producer_commit();
pipeline.producer_acquire(); // This acquire always blocks, deadlocks, ... ??
...
```

The second `producer_acquire` call cannot ever make progress AFAICT. And that's something that we also guarantee.

Not sure where would be the right place to document this, but I think it should be in the libcudacxx docs.

c0riolis: Yes, let's get this change in, we can iterate on top of it after.

c0riolis: Created #77 for this.

griwes (approved):

gonzalobg:
Thank you @c0riolis for doing all this work, the new docs look great!

PR #56, opened 2020-10-29
Fix <tuple> on MSVC

wmaxey:
`decltype` is a culprit for a slew of MSVC bugs. About 50 failures have been fixed by hacking the `__tuple_sfinae_base` trait.

Tests that still need to be addressed:
```
Failing Tests (8):
    libcu++ :: std/utilities/tuple/tuple.tuple/tuple.apply/apply_extended_types.pass.cpp
    libcu++ :: std/utilities/tuple/tuple.tuple/tuple.cnstr/PR27684_contains_ref_to_incomplete_type.pass.cpp
    libcu++ :: std/utilities/tuple/tuple.tuple/tuple.cnstr/UTypes.pass.cpp
    libcu++ :: std/utilities/tuple/tuple.tuple/tuple.cnstr/alloc.pass.cpp
    libcu++ :: std/utilities/tuple/tuple.tuple/tuple.cnstr/deduct.pass.cpp
    libcu++ :: std/utilities/tuple/tuple.tuple/tuple.cnstr/nothrow_cnstr.pass.cpp
    libcu++ :: std/utilities/tuple/tuple.tuple/tuple.cnstr/test_lazy_sfinae.pass.cpp
    libcu++ :: std/utilities/tuple/tuple.tuple/tuple.helper/tuple_size_structured_bindings.pass.cpp

  Expected Passes    : 70
  Unexpected Failures: 8
```


wmaxey:
After a few touch ups MSVC builds clean with only four tests unsupported.

Testing Time: 136.69s
  Expected Passes    : 76
  Unsupported Tests  : 4

two of these are actually ICEs or MSVC not supporting something.

Incomplete types for example cannot be used with some intrinsics.

.upstream-tests/test/std/utilities/tuple/tuple.tuple/tuple.cnstr/UTypes.pass.cpp
```diff
@@ -139,9 +139,9 @@ int main(int, char**)
 
         Tup t2(E(0));
         static_assert(!test_convertible<Tup, E>(), "");
-        assert(cuda::std::get<0>(t) == 0);
```
griwes, line 142: ...that's embarrassing.

libcxx/include/__tuple
```diff
@@ -254,29 +254,30 @@ template <class ..._Tp> struct __tuple_types {};
 
 namespace __indexer_detail {
 
-template <size_t _Idx, class _Tp>
-struct __indexed { using type _LIBCUDACXX_NODEBUG_TYPE = _Tp; };
+template <size_t _Idx, class ..._Types>
+struct _one_of;
```
griwes, line 258: This would be better off being called `_nth_of`, right?

libcxx/include/tuple
```diff
@@ -382,8 +382,23 @@ struct _LIBCUDACXX_DECLSPEC_EMPTY_BASES __tuple_impl<__tuple_indices<_Indx...>,
     _LIBCUDACXX_CONSTEXPR __tuple_impl()
         _NOEXCEPT_(__all<is_nothrow_default_constructible<_Tp>::value...>::value) {}
 
+    // Handle non-allocator, full initialization
     template <size_t ..._Uf, class ..._Tf,
-              size_t ..._Ul, class ..._Tl, class ..._Up>
+              size_t ..._Ul, class ..._Tl, class ..._Up,
+              _EnableIf<sizeof...(_Uf) == sizeof...(_Tp), bool> = false>
```
griwes, line 388: I know the pack sizes are necessarily the same, but... how do would you feel about comparing `_Tf` with `_Tp`, instead of `_Uf`?

libcxx/include/tuple
```diff
@@ -524,22 +588,19 @@ class _LIBCUDACXX_TEMPLATE_VIS tuple
                 __tuple_constructible<
                     tuple<_Args...>,
                     typename __make_tuple_types<tuple,
-                             sizeof...(_Args) < sizeof...(_Tp) ?
-                                 sizeof...(_Args) :
-                                 sizeof...(_Tp)>::type
+                            __lesser<sizeof...(_Args), sizeof...(_Tp)>::value
```
griwes, line 591: Is this necessary for some reason, or just a convenience rewrite? The new version will probably be more expensive, because it now has to instantiate a type for every combination of `_Lh` and `_Rh`, while the old one just compared values, so if this isn't a workaround for something, we should probably go back to the original way this was written.

libcxx/include/__tuple
```diff
@@ -492,63 +495,6 @@ struct _LIBCUDACXX_TYPE_VIS __check_tuple_constructor_fail {
...
-template <>
-struct __sfinae_assign_base<false, true> {
-  __sfinae_assign_base() = default;
-  __sfinae_assign_base(__sfinae_assign_base const&) = default;
-  __sfinae_assign_base(__sfinae_assign_base &&) = default;
-  __sfinae_assign_base& operator=(__sfinae_assign_base const&) = delete;
-  __sfinae_assign_base& operator=(__sfinae_assign_base&&) = default;
-};
-#endif // _LIBCUDACXX_STD_VER > 14
-
```
griwes, line 551: So... where did this functionality go?

libcxx/include/__tuple
wmaxey: Sure, I can rename it to that.

libcxx/include/tuple
wmaxey: It was for convenience, but can be reverted at this point. When I was nursing that `__indexed` type error I was trying to get rid of the ternary to see if it helped.

libcxx/include/__tuple
wmaxey: It was dead code. I had rewritten some parts of the base types and had left it in on accident when I reverted it.

libcxx/include/tuple
wmaxey: Is there a reason to?

wmaxey: Ah wait, I just noticed that it's different, but the value would be the same still...

griwes: It'd then compare the sizes of the type packs, instead of comparing the size of an index pack with a type pack. It doesn't matter semantically; I just wonder if it'd read more nicely with such a change, since it stood out a tiny bit to me. Feel free to disregard this though ;)

wmaxey:
Internal CI run on CL 29115815

griwes (approved):
This looks good to me. Have you tested with NVRTC?

wmaxey:
A fresh internal CI run includes an NVRTC pass.

wmaxey:
NVRTC is clean on local builds, however I picked this up after updating to latest available GCC: #67 

I'll see if I can submit a fix for this later. It's not related to this change for the time being and breaks on main.

wmaxey:
CI showed that I missed the case for when `std::tuple<std::array<type, size>>` can be used to create a tuple made up of n-types. I've updated the offending commit and have started re-running CI.

.upstream-tests/test/std/utilities/tuple/tuple.tuple/tuple.cnstr/deduct.pass.cpp
```diff
@@ -14,7 +14,7 @@
 // GCC's implementation of class template deduction is still immature and runs
 // into issues with libc++. However GCC accepts this code when compiling
 // against libstdc++.
-// XFAIL: gcc-5, gcc-6, gcc-7
+// XFAIL: gcc-5, gcc-6, gcc-7, gcc-10.2
```
wmaxey, line 17: See issue #67 about this failure.

It could be resolvable with some investigation since earlier versions are passing.

wmaxey:
CI is clean after marking the last few failures as expected or having fixed them. I'll go ahead and merge this.

CL: 29308251

PR #69, opened 2020-11-03
Barrier parity waiting and algorithm tweaks

ogiroux:
Adds the ability to wait on a barrier when having only retained phase parity information (from domain knowledge, it isn't recoverable from the barrier object or its tokens), like so:

```c++
barrier<> b (1);
(void)b.arrive(); // discard the nodiscard return
barrier_wait_for_parity(&b, false);
```

Also changed the algorithm so that some `b.arrive();` can be equivalent to just a `fetch_add` on its constituent atomic. These are available with the `__expect_extra_arrive` and `__extra_arrive` members.

The parity waiting part isn't ABI breaking, but the algorithm tweak is ABI breaking because it halves the value of `::max()`, to half of int max (_which is somewhat theoretical an issue given how large this number is_). I didn't add ABI handling code.

### DIFF
--- a/include/cuda/std/barrier
+++ b/include/cuda/std/barrier
@@ -285,7 +285,7 @@ public:
     }
 
     inline _LIBCUDACXX_INLINE_VISIBILITY
-    friend void wait_for_parity(barrier const* __self, bool __parity)
+    friend void barrier_wait_for_parity(barrier const* __self, bool __parity)
     {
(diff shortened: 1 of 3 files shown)


ogiroux:
I amended the description above after reducing the ABI exposure.

brycelelbach:
Closing in favor of #111

PR #74, opened 2020-11-16
Added some wording for the init function.

ogiroux:


docs/api/synchronization_library/barrier.md
```diff
@@ -23,26 +23,35 @@ cuda::std::barrier<> bb;
 cuda::barrier<cuda::thread_scope_block> c;
 ```
 
-The class template `barrier` may also be declared without initialization; a
+The class template `barrier` may also be declared without initialization in the `cuda::` namespace; a
 friend function `init` may be used to initialize the object.
```
gonzalobg, line 27:
```suggestion
Variable declarations whose type is an instantiation of the `barrier` class template may be initialized with the friend function `init`.
```

I think the "class template `barrier`" is already declared in the `::cuda` namespace and cannot be re-declared there. What can be declared without initialization are variables whose type is an instantiation of that template.

docs/api/synchronization_library/barrier.md
```diff
@@ -23,26 +23,35 @@ cuda::std::barrier<> bb;
 cuda::barrier<cuda::thread_scope_block> c;
```
gonzalobg, line 23: What does this default/value/trivial initialization do for barrier? 

`std::barrier` only has the constructor `barrierbarrier(ptrdiff_t expected, CompletionFunction f = CompletionFunction());`

What is the `expected` count set to when using default/value/trivial initialization?

docs/api/synchronization_library/barrier.md
```diff
@@ -23,26 +23,35 @@ cuda::std::barrier<> bb;
...
 ```
 
-The class template `barrier` may also be declared without initialization; a
+The class template `barrier` may also be declared without initialization in the `cuda::` namespace; a
 friend function `init` may be used to initialize the object.
 
 ```c++
 // Shared memory does not allow initialization.
 __shared__ cuda::barrier<cuda::thread_scope_block> b;
```
gonzalobg, line 31: I don't know of a specification of the exact semantics of `__shared__` in CUDA C++ and how they interact with default constructors, but I imagine they are very similar to obtaining a pointer to a raw allocation with kernel lifetime, e.g., 

```c++
cuda::barrier<cuda::thread_scope_block>* b = ...points to uninitialized storage...;
```

where `init(b, 1)` acts like `new(b) cuda::barrier<cuda::thread_scope_block>{1}`.

Placement-new guarantees that the object will be initialized in place. Why do we need `init` ? Couldn't we implement the constructor as follows:

```c++
cuda::barrier<cuda::thread_scope_block>::barrier(ptrdiff_t expected) {
    init(this, expected);
}
```

and just use "placement `new`" for initializing barriers?

docs/api/synchronization_library/barrier.md
```diff
@@ -23,26 +23,35 @@ cuda::std::barrier<> bb;
...
+namespace cuda {
+  template<thread_scope Sco, class CompletionF>
+  __host__ __device__ void init(barrier<Sco,CompletionF>* bar, std::ptrdiff_t expected);
+  template<thread_scope Sco, class CompletionF>
+  __host__ __device__ void init(barrier<Sco,CompletionF>* bar, std::ptrdiff_t expected, CompletionF completion);
+}
+*/
 ```
 
+- Expects: `*bar` is trivially initialized.
```
gonzalobg, line 44: If the intent is to exhibit undefined behavior if this condition is not met: 

```suggestion
- Preconditions: `*bar` is trivially initialized.
```

See: https://eel.is/c++draft/structure.specifications#3

docs/api/synchronization_library/barrier.md
```diff
@@ -23,26 +23,35 @@ cuda::std::barrier<> bb;
...
+- Effects: equivalent to initializing `*bar` with a constructor.
```
gonzalobg, line 45: Which constructor? Arguable there is only one:

```suggestion
- Effects: equivalent to initializing `*bar` with `barrier::barrier(expected, completion)`.
```

docs/api/synchronization_library/barrier.md
```diff
@@ -23,26 +23,35 @@ cuda::std::barrier<> bb;
...
 // Shared memory does not allow initialization.
 __shared__ cuda::barrier<cuda::thread_scope_block> b;
 
 init(&b, 1); // Use this friend function to initialize the object.
+/*
+namespace cuda {
+  template<thread_scope Sco, class CompletionF>
+  __host__ __device__ void init(barrier<Sco,CompletionF>* bar, std::ptrdiff_t expected);
+  template<thread_scope Sco, class CompletionF>
+  __host__ __device__ void init(barrier<Sco,CompletionF>* bar, std::ptrdiff_t expected, CompletionF completion);
```
gonzalobg, line 39: Why do we add two overloads of `init` instead of one with a default `CompletionF` ? 

```suggestion
  template<thread_scope Sco, class CompletionF>
  __host__ __device__ void init(barrier<Sco,CompletionF>* bar, std::ptrdiff_t expected, CompletionF completion = CompletionF());
```

docs/api/synchronization_library/barrier.md
griwes:
```suggestion
Variables declared with a type that is an instantiation of the `cuda::barrier` class template may be declared without initialization; in such case, they must be initialized by calling a friend function `init` prior to any use.
```

How about the above?

docs/api/synchronization_library/barrier.md
griwes: That is more or less what happens; but having non-experts write placement new expressions is probably not the greatest idea.

docs/api/synchronization_library/barrier.md
griwes: I agree with Gonzalo that we should *document* this with just a single overload with `completion` defaulted.

The reason for why this is *implemented* as two overloads is that the friend functions are instantiated when you instantiate the class template, so that they can be found by ADL; and if `CompletionF` does not support default construction, the default argument is ill-formed and the entire program fails (even when you don't actually call `init`, if I remember how this works and what I encountered here correctly).

gonzalobg: That makes sense. 

The standard requires the default type of `CompletionF` to be `DefaultConstructible`, http://eel.is/c++draft/thread.barrier#class-6 , but it allows users to specify a non-`DefaultConstructible` one.

docs/api/synchronization_library/barrier.md
griwes: Looking at my suggested change now, I think I'll want to tweak further, but that's a thing for tomorrow.

brycelelbach:
This will need a rebase.

miscco:
@griwes would you mind pushing this over the finishing line?

PR #76, opened 2020-11-17
Enable `is_constant_evaluated` in c++11

wmaxey:
This seems acceptable to backport and will be useful for constexpr `<complex>`.

We test on the builtin being available, however, in some compiler versions it won't be depending on host+NVCC combination. 

We should also look into exposing feature test macros with some mechanism similar to `<version>`


jrhemstad (approved):

wmaxey:
CI is clean and fixes several old failures. 29334943

PR #79, opened 2020-11-19
MSVC: Fix <atomic> for versions older than 19.24 by using _InterlockedExchange and implicit volatile read behavior

wmaxey:


libcxx/include/support/win32/atomic_msvc.h
```diff
@@ -76,25 +93,47 @@ void __atomic_load(const volatile _Type *__ptr, _Type *__ret, int __memorder) {
...
 template<class _Type, detail::_enable_if_sized_as<_Type,4> = 0>
-void __atomic_store_relaxed(const volatile _Type *__ptr, _Type *__val) {
-    __int32 __tmp = reinterpret_cast<__int32&>(*__val);
-    __iso_volatile_store32((volatile __int32 *)__ptr, __tmp);
+void __atomic_store_relaxed(volatile _Type *__ptr, _Type *__val) {
+    auto __t = reinterpret_cast<__int32 *>(__val);
+    auto __d = reinterpret_cast<volatile __int32 *>(__ptr);
+#ifdef _LIBCUDACXX_MSVC_HAS_NO_ISO_INTRIN
+    // int cannot be converted to long?...
+    (void)_InterlockedExchange(reinterpret_cast<volatile long *>(__d), *__t);
```
wmaxey, line 121: MSVC wouldn't accept an implicit cast to `long`.

I don't think we need any explicit guarantees that long is always going to be 32 bits on MSVC.

griwes (approved):

wmaxey:
29339459.1 Actually runs clean on 14.14 and 14.20. Local runs of more recent versions run clean as well.

PR #123, opened 2021-02-23
Rename the is_convertible fallback helper to avoid a clash with a keyword.

### LINKED ISSUE #118: `__is_convertible` is used as a library name, but is also a context sensitive keyword, causing EDG complaints
@griwes believes the best fix is to just rename our `__is_convertible` to something else. It only appears on a fallback path that we currently don't take.

This is NVBug 3252533

griwes:
NVCC has changed the behavior of Clang keywords like __is_convertible, so we
need to adjust.

Bug 3252533
Fixes #118

Running internal CI on CL 29638598.


wmaxey:
I'm happy with the `fallback` suffix, approving

wmaxey (approved):

PR #143, opened 2021-03-19
Document interaction of asynchronous operations with thread scopes

### LINKED ISSUE #137: Document why combining pipeline<thread_scope_thread>/barrier<thread_scope_thread> with memcpy_async is legal
According to the thread scope documentation:

> A thread scope specifies the kind of threads that can synchronize with each other using a primitive such as an atomic or a barrier.
>
> Each thread (CPU or GPU) is related to itself by the thread thread scope, specified with thread_scope_thread.

[This example](https://nvidia.github.io/libcudacxx/extended_api/synchronization_primitives/pipeline_consumer_wait_prior.html#example), however, suggests that multiple threads of execution - the current thread and the threads performing the `memcpy_async` - can be synchronized using the synchronization primitives with `thread_scope_thread`. 

A fix could be to change the documentation to:

> The thread thread scope relates each thread (CPU or GPU) and its `???` threads to itself.

where `???` would be a name for the set of threads that includes the ones used by the `memcpy_async`s.

gonzalobg:
Closes #137.

### DIFF
--- a/docs/extended_api/asynchronous_operations.md
+++ b/docs/extended_api/asynchronous_operations.md
@@ -1,7 +1,13 @@
 ## Asynchronous Operations
 
+Asynchronous operations are performed _as-if_ in other threads. This other threads are related to the thread requesting the asynchronous operations via all scope relationships [^thread_scope].
+
+
+
 | [`cuda::memcpy_async`] | Asynchronously copies one range to another. `(function template)` <br/><br/> 1.1.0 / CUDA 11.0 <br/> 1.2.0 / CUDA 11.1 (group & aligned overloads) |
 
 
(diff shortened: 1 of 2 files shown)


docs/extended_api/asynchronous_operations.md
```diff
@@ -1,7 +1,13 @@
 ## Asynchronous Operations
 
+Asynchronous operations are performed _as-if_ in other threads. This other threads are related to the thread requesting the asynchronous operations via all scope relationships [^thread_scope].
```
gonzalobg, line 3:
```suggestion
Asynchronous operations are potentially performed in a new thread. This new threads are related to the thread requesting the asynchronous operations via all scope relationships [^thread_scope].
```

gonzalobg:
See #371

PR #156, opened 2021-04-19
docker configuration for ppc64le gcc8/clang11/hpc20.9

veictry:
create ppc64le docker configuration. 

all base os is moved to centos now.


wmaxey:
Are there any DVS virtuals or verification that the new configurations work?

The additions look good to me, but I don't have the ability to test them.

miscco:
@wmaxey Is this something we still want given the current blossom infra we have?

PR #161, opened 2021-04-27
Align complex types to total size of both components

wmaxey:


griwes (changes requested):
The alignment must be conditional on the abi version. See https://github.com/NVIDIA/libcudacxx/blob/4b1a1df666f3bac26f60cb06cda051ca6fc9b158/libcxx/include/barrier#L200 for prior art.

griwes:
Wouldn't a simpler change of just slapping `alignas(sizeof(T))` on complex itself work?

griwes:
This also needs tests that verify the alignments for both the old abi versions and the new one; there's a barrier v2 abi test in heterogeneous tests that you can use as prior art for how to approach it in general.

wmaxey:
> Wouldn't a simpler change of just slapping `alignas(2*sizeof(T))` on complex itself work?

Funny, I hadn't thought of that. Originally I had wrapped the members in anonymous structs, but NVRTC doesn't support that extension.

wmaxey:
* Simplified implementation
* Added ABI version tests
* Currently passes on GCC, Clang, and NVRTC builds
  * MSVC TBD

wmaxey:
Closing as #172 will supersede this PR.

PR #164, opened 2021-05-13
use python3 to avoid some syntax issue

veictry:
new lit is released, but there is a syntax issue when we are using new lit tool

use Python3 can solve this issue:

http://scbuilds4u.nvidia.com/dvs/#/change/2995837939432407.1?showTab=DVS

the failures in this virtual is caused by NVCC compiler issue.


wmaxey (approved):

wmaxey:
I will submit this to //sw/gpgpu when merged.

PR #165, opened 2021-05-17
Fix NVCC test failures involving long double

wmaxey:
This patch fixes NVCC's complaints about using long double with some type checks:

```C++
//    Make sure the types are right
    static_assert ( cuda::std::is_same<decltype( 3h   ), cuda::std::chrono::hours>::value, "" );
    static_assert ( cuda::std::is_same<decltype( 3min ), cuda::std::chrono::minutes>::value, "" );
    static_assert ( cuda::std::is_same<decltype( 3s   ), cuda::std::chrono::seconds>::value, "" );
    static_assert ( cuda::std::is_same<decltype( 3ms  ), cuda::std::chrono::milliseconds>::value, "" );
    static_assert ( cuda::std::is_same<decltype( 3us  ), cuda::std::chrono::microseconds>::value, "" );
    static_assert ( cuda::std::is_same<decltype( 3ns  ), cuda::std::chrono::nanoseconds>::value, "" );
```

Original error:

```
2021-05-17 19:15:51+00:00| Error #20208-D: 'long double' is treated as 'double' in device code
2021-05-17 19:15:51+00:00| 
2021-05-17 19:15:51+00:00| Error #20208-D: 'long double' is treated as 'double' in device code
2021-05-17 19:15:51+00:00| 
2021-05-17 19:15:51+00:00| Error #20208-D: 'long double' is treated as 'double' in device code
2021-05-17 19:15:51+00:00| 
2021-05-17 19:15:51+00:00| Error #20208-D: 'long double' is treated as 'double' in device code
2021-05-17 19:15:51+00:00| 
2021-05-17 19:15:51+00:00| Error #20208-D: 'long double' is treated as 'double' in device code
2021-05-17 19:15:51+00:00| 
2021-05-17 19:15:51+00:00| Error #20208-D: 'long double' is treated as 'double' in device code
2021-05-17 19:15:51+00:00| 
2021-05-17 19:15:51+00:00| 6 errors detected in the compilation of "/sw/gpgpu/libcudacxx/.upstream-tests/test/std/utilities/time/time.duration/time.duration.literals/literals2.pass.cpp".
2021-05-17 19:15:51+00:00| # --error 0x1 --
```

According to the draft for chrono literals, the duration representation is unspecified, so this should be fine, and makes NVCC happy.
http://eel.is/c++draft/time.duration.literals

### DIFF
--- a/libcxx/include/chrono
+++ b/libcxx/include/chrono
@@ -1013,7 +1013,7 @@ round(const duration<_Rep, _Period>& __d)
         return __upper;
     return __lower.count() & 1 ? __upper : __lower;
 }
-#endif // _LIBCUDACXX_STD_VER > 11 
+#endif // _LIBCUDACXX_STD_VER > 11
 
 // duration
 
@@ -2539,7 +2539,7 @@ inline constexpr days year_month_day::__to_days() const noexcept
     static_assert(std::numeric_limits<unsigned>::digits >= 18, "");
     static_assert(std::numeric_limits<int>::digits >= 20     , "");
 
-    // nvcc doesn't allow ODR using constexpr globals. Therefore, 
+    // nvcc doesn't allow ODR using constexpr globals. Therefore,
     // make a temporary initialized from the global
     auto constexpr __Feb = February;
     const int      __yr  = static_cast<int>(__y) - (__m <= __Feb);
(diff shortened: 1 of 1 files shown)


griwes:
@brycelelbach this is an abi break, but probably one we don't care about. Do you agree with this assessment?

brycelelbach:
It doesn't seem like a big deal, but it is an ABI break. We are reving the ABI in 2.0.0 for `complex` anyways, so we might as well hold this off until then.

So I would:

* First, disable the failing parts of the test, then integrate that to 1.6.0.
* Then, I would fix it by making these double instead of long double and target that for 2.0.0

But keep in mind user defined literals must take a `long double`.

brycelelbach:
This looks good.

brycelelbach:
Actually, maybe we move the ABI break from 2.0.0 to 1.6.0.

wmaxey:
Closing as this change is included in #172.

PR #166, opened 2021-05-18
Docker: Use older version of lit on Ubuntu versions where it fails to run

wmaxey:


wmaxey:
A few lines may have been missed in #164 

VDVS: http://ausdvs.nvidia.com/Build_Results?virtualId=1000905067&which_page=current_build

brycelelbach (approved):

PR #177, opened 2021-07-01
Make <nv/target> usable in C and C++98 compilations

dkolsen-pgi:
The features of `<nv/target>` are only available in C++11 CUDA mode.  But some headers that use `<nv/target>` can be compiled, with limited functionality, as C or as C++98.  `<nv/target>` needs to do the same.

This change allows `<nv/target>` to be included successfully when compiling as C or as C++98.  In this situation, "`if target`" won't
work because the namespace `nv::target` won't exist, and the compatibility macros will assume host-only.

### DIFF
--- a/include/nv/target
+++ b/include/nv/target
@@ -15,7 +15,7 @@
 
 #if defined(__NVCC__) || defined(__CUDACC_RTC__)
 #  define _NV_COMPILER_NVCC
-#elif defined(__NVCOMPILER)
+#elif defined(__NVCOMPILER) && __cplusplus >= 201103L
 #  define _NV_COMPILER_NVCXX
 #else
 #endif
@@ -25,6 +25,8 @@
 #else
 #  define _NV_BITSET_ATTRIBUTE
 #endif
+
+#if __cplusplus >= 201103L
 
 namespace nv {
   namespace target {
@@ -163,6 +165,8 @@ namespace nv {
     using detail::provides;
   }
 }
+
+#endif // C++11
 
 #include "detail/__target_macros"
 


wmaxey:
We have no configured C++03 testing or compiler configurations, were you able to run this on NVC++ and others yourself?

include/nv/target
```diff
@@ -26,6 +26,8 @@
 #  define _NV_BITSET_ATTRIBUTE
 #endif
 
+#if __cplusplus >= 201103L
```
wmaxey, line 29: I would like to see instead:
```
#if c++11
# define _NV_CONSTEXPR_11
#endif
...

#if _NV_CONSTEXPR_11
... <- nv::target:: stuff
#endif
```

But I can address this myself later. I won't hold up the commit on this.

dkolsen-pgi: `constexpr` isn't the only C++11 feature in there.  There's also an `enum class`.  So I'm a little reluctant to use a macro that focuses on the `constexpr`.

dkolsen-pgi:
I manually tested the change by compiling this program every way I could, with and without CUDA turned on, with `nvcc`, `nvc++`, and `gcc`, and as C, C++98, C++11, and C++14.

```
#include <stdio.h>
#include <nv/target>
int main() {
  NV_IF_TARGET(NV_IS_HOST, (
    printf("Passed\n");
  ), (
    printf("FAILED: This block should never be run.\n");
  ));
}
```

wmaxey (approved):

wmaxey:
Rebased and launched virtual, I will submit when clean.

http://scdvs.nvidia.com/Build_Results?virtualId=1000072758&which_page=current_build

PR #179, opened 2021-07-07
Refactor <atomic> and move implementation to libcxx

wmaxey:
# Overview

This is a refactor of `<cuda/std/atomic>` and the underlying `<atomic>` headers. This moves the implementation of CUDA specific atomics to the inner implementation. It also allows for us to implement the underlying mechancis for `atomic_ref` in the near future and allows NVC++ to access atomic PTX intrinsics when support is finally enabled.

# Motivation

There are several goals here:
   - Eases the review burden on the nvcxx feature branch.
   - An opportunity to fix some alignment bugs with `atomic<struct>` types. 
   - Allows us to prepare and test a backend that supports `atomic_ref`, a feature we will support very soon.

# Requirements

 - Begin using `<nv/target>` in the atomic backend and within relevant tests.
 - Move atomic intrinsics from <cuda/std/...> and instead into `<libcxx/support/atomic>.
 - Refactor the `atomic_impl` internals to support more generic host and device layering.

# Design

 - Move implementation specific atomic backends into individual headers.
 - When the CUDA backend is enabled `#include` a relevant host atomic backend into a `host::` namespace.
 - For `atomic_ref` implement a corresponding atomic implementation class that is compatible with `__cxx_atomic_unwrap`.
 - Remove C11 support or continue to support it in host only code.

# Test Plan

 - All currently existing tests should continue to function as expected.
 - Add a test for ensuring the correctness of differently aligned types.

## Performance Tests

 - Compare generated code to previously generated code. The results of this should indicate that nothing will have changed.

# Documentation

 - No documentation changes needed as of yet.


.upstream-tests/test/cuda/pipeline_memcpy_async_thread_scope_generic.h
```diff
@@ -107,19 +107,23 @@ __host__ __device__ __noinline__
...
 {
     test_select_destination<Scope, T, local_memory_selector>();
-#ifdef __CUDA_ARCH__
-    test_select_destination<Scope, T, shared_memory_selector>();
-    test_select_destination<Scope, T, global_memory_selector>();
-#endif
+    NV_DISPATCH_TARGET(
+        NV_IS_DEVICE, (
+            // test_select_destination<Scope, T, shared_memory_selector>();
+            // test_select_destination<Scope, T, global_memory_selector>();
```
griwes, line 126: Is this a leftover from something?

.upstream-tests/test/std/atomics/atomics.flag/default.pass.cpp
```diff
@@ -25,26 +25,40 @@
...
+                A& zero = *new (storage) A();
+                assert(!zero.test_and_set());
+                _TEST_NO_DESTRUCT_ZERO;
+            ),
+            NV_IS_HOST, (
+                typedef cuda::std::atomic_flag A;
+                TEST_ALIGNAS_TYPE(A) char storage[sizeof(A)] = {1};
+                A& zero = *new (storage) A();
+                assert(!zero.test_and_set());
+                _TEST_NO_DESTRUCT_ZERO;
```
griwes, line 59: Move into a lambda/named function and invoke that lambda/function on the branches to avoid code duplication?

.upstream-tests/test/support/concurrent_agents.h
```diff
@@ -11,16 +11,15 @@
     #include <thread>
 #endif
 
+#if defined(__CUDA_MINIMUM_ARCH__) && __CUDA_MINIMUM_ARCH__ < 350
+    #error "This test requires CUDA dynamic parallelism to work."
+#endif
+
```
griwes, line 17: This hasn't been true in a while, prune.

.upstream-tests/test/support/cuda_space_selector.h
```diff
@@ -36,6 +39,20 @@ struct malloc_memory_provider {
...
+    }
+
+
+#if !defined(__CUDACC_RTC__)
+    __host__ char* host_static_storage() {
+        alignas(T*) static char storage[shared_offset];
+        return storage;
+    }
+#endif
+
```
griwes, line 55: Where is this used?

.upstream-tests/test/support/cuda_space_selector.h
```diff
@@ -165,19 +182,30 @@ class memory_selector
         return ptr;
     }
```
griwes, line 183: Does this not explode on NVC++?

griwes (commented):
Partial review, only tests so far.

.upstream-tests/test/cuda/pipeline_memcpy_async_thread_scope_generic.h
wmaxey: Yeah, I'm on the lookout for more of these. This was from some earlier testing on NVC++.

wmaxey:
> Partial review, only tests so far.

I've removed test changes in the interest of time.

Unfortunately there is a bug blocking progress related to `<nv/target>`: https://nvbugs/3341536

libcxx/include/support/atomic/atomic_cuda.h
```diff
@@ -0,0 +1,488 @@
...
+
+template <int _Scope>
+_LIBCUDACXX_INLINE_VISIBILITY auto constexpr __scope_tag() ->
+        typename __scope_enum_to_tag<_Scope>::type {
+    return typename __scope_enum_to_tag<_Scope>::type();
+}
+// END TODO
+
+// Wrap host atomic implementations into a sub-namespace
+namespace host {
```
griwes, line 96: `__host`

libcxx/include/support/atomic/atomic_cuda.h
```diff
@@ -0,0 +1,488 @@
...
+//TODO
+// #  include "atomic_c11.h"
+#endif
+}
+
+#include "atomic_cuda_generated.h"
+#include "atomic_cuda_derived.h"
+
+template <typename _Tp>
+struct __skip_amt { enum {value = 1}; };
```
griwes, line 111: "amt"? "amount"? I understand what this is for because I know the __atomic intrinsics, but I'd like this to have a better name. "__difference_scale" or something? (I don't think that's a good name either, but it's more evocative of what it is used for.)

libcxx/include/support/atomic/atomic_cuda.h
```diff
@@ -0,0 +1,488 @@
...
+    return __cxx_small_from_32<_Tp>(__cxx_atomic_load(&__a->__a_value, __order));
+}
+
+template <typename _Tp, int _Sco>
+__host__ __device__ inline _Tp __cxx_atomic_exchange(__cxx_atomic_base_small_impl<_Tp, _Sco> volatile* __a, _Tp __value, memory_order __order) {
+    return __cxx_small_from_32<_Tp>(__cxx_atomic_exchange(&__a->__a_value, __cxx_small_to_32(__value), __order));
+}
+__host__ __device__
+inline int __cuda_memcmp(void const * __lhs, void const * __rhs, size_t __count) {
+#ifdef __CUDA_ARCH__
```
griwes, line 425: Do we want to give this `if target` treatment? If not, why?

libcxx/include/support/atomic/atomic_cuda_generated.h
```diff
@@ -117,11 +114,11 @@ __device__ void __atomic_store_cuda(volatile _Type *__ptr, _Type *__val, int __m
...
-template<class _CUDA_A, class _CUDA_B, class _CUDA_C, class _CUDA_D> static inline __device__ void __cuda_compare_exchange_release_32_block(_CUDA_A __ptr, _CUDA_B& __dst, _CUDA_C _cmp, _CUDA_D __op) { asm volatile("atom.cas.release.cta.b32 %0,[%1],%2,%3;" : "=r"(__dst) : "l"(__ptr),"r"(_cmp),"r"(__op) : "memory"); }
-template<class _CUDA_A, class _CUDA_B, class _CUDA_C, class _CUDA_D> static inline __device__ void __cuda_compare_exchange_volatile_32_block(_CUDA_A __ptr, _CUDA_B& __dst, _CUDA_C _cmp, _CUDA_D __op) { asm volatile("atom.cas.cta.b32 %0,[%1],%2,%3;" : "=r"(__dst) : "l"(__ptr),"r"(_cmp),"r"(__op) : "memory"); }
+template<class _CUDA_A, class _CUDA_B, class _CUDA_C, class _CUDA_D> static inline __device__ void __cuda_compare_exchange_acq_rel_32_block(_CUDA_A __ptr, _CUDA_B& __dst, _CUDA_C __cmp, _CUDA_D __op) { asm volatile("atom.cas.acq_rel.cta.b32 %0,[%1],%2,%3;" : "=r"(__dst) : "l"(__ptr),"r"(__cmp),"r"(__op) : "memory"); }
```
griwes, line 117: Fix this in the generator too.

libcxx/include/support/atomic/atomic_msvc.h
```diff
@@ -27,7 +27,37 @@
...
+    int const __max = __a > __b ? __a : __b;
+    if(__max != __ATOMIC_RELEASE)
+        return __max;
+    static int const __xform[] = {
+        __ATOMIC_RELEASE,
+        __ATOMIC_ACQ_REL,
+        __ATOMIC_ACQ_REL,
+        __ATOMIC_RELEASE };
+    return __xform[__a < __b ? __a : __b];
+}
```
griwes, line 60: Can we extract these up a layer to not repeat them?

libcxx/include/support/atomic/atomic_msvc.h
```diff
@@ -27,7 +27,37 @@
     #error Unsupported hardware
 #endif // hardware
 
-_LIBCUDACXX_BEGIN_NAMESPACE_CUDA
+_LIBCUDACXX_INLINE_VISIBILITY inline _LIBCUDACXX_CONSTEXPR int __to_gcc_order(memory_order __order) {
```
griwes, line 30: I understand where the name comes from, but it's a bit awkward to use "gcc" in it for MSVC. We should figure out a better name for those macros than "gcc order".

libcxx/include/support/atomic/atomic_msvc.h
```diff
@@ -435,4 +463,213 @@ _Type __host__ __atomic_fetch_min(_Type volatile *__ptr, _Delta __val, int __mem
     return __expected;
 }
 
-_LIBCUDACXX_END_NAMESPACE_CUDA
+template <typename _Tp, int _Sco>
```
griwes, line 466: The code from here onwards repeats for GCC and MSVC, right? Can we move it to some sort of a "derived" header that is just included in both of these headers to avoid repeating all of it?

Maybe if we had a from-scratch backend for MSVC this wouldn't be an issue, but seems we aren't really in such a place, right?

wmaxey: It would be 'easiest' to just strip the content from `atomic_gcc.h` and instead call that some kind of `cxx_atomic_base.h`

`atomic_gcc.h` need only include `cxx_atomic_base.h` since gcc/clang comes with the intrinsics.

`atomic_msvc.h` creates the back-end intrinsics and then includes `cxx_atomic_base.h`.

Thoughts on that method? I think that would accomplish reducing code duplication and wouldn't be too confusing. Though it still means everything relies on the GCC atomic symbols being defined.

libcxx/include/support/atomic/atomic_cuda_generated.h
wmaxey: That's the weird thing. This is already in the generator. :/ A switch from `__cmp` to `_cmp` had snuck into an earlier change, but they are synchronized now.

libcxx/include/support/atomic/atomic_cuda.h
wmaxey: This is essentially the same hack that was in another layer. I'll try to use that one instead and get rid of the duplication.

libcxx/include/support/atomic/atomic_cuda.h
wmaxey: Will uglify!

libcxx/include/support/atomic/atomic_msvc.h
wmaxey: Fixed as part of the dedup commit. Still not renamed, though.

libcxx/include/support/atomic/atomic_cuda.h
wmaxey: This should be fixed now.

include/cuda/std/atomic
```diff
@@ -46,16 +47,26 @@
...
 
 _LIBCUDACXX_BEGIN_NAMESPACE_CUDA
 
+using std::detail::thread_scope;
+using std::detail::thread_scope_system;
+using std::detail::thread_scope_device;
+using std::detail::thread_scope_block;
+using std::detail::thread_scope_thread;
+
+namespace detail {
```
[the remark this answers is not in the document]
wmaxey: Fixed in a rename commit.

libcxx/include/support/atomic/atomic_cuda.h
wmaxey: Fixed with a rename to __atomic_ptr_inc. I think that conveys the idea that it's an increment for pointers.

wmaxey:
SC-DVS: https://scbuilds4u/dvs/#/change/3025360339432407.2?eventType=Virtual
AUS-DVS: https://builds4u.nvidia.com/dvs/#/change/3025362855810995.1?eventType=Virtual&dvs_show=sanity

include/cuda/std/barrier
```diff
@@ -293,7 +293,7 @@ inline void __strided_memcpy(char * __destination, char const * __source, std::s
     }
 }
 
-#if __CUDA_ARCH__ >= 800
+#if __CUDA_MINIMUM_ARCH__ >= 800
```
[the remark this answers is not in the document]
wmaxey: Fixed remaining __CUDA_ARCH__ dispatches.

libcxx/include/support/atomic/atomic_msvc.h
wmaxey: Renamed as something more generic.

libcxx/include/support/atomic/atomic_msvc.h
wmaxey: Fixed.

include/cuda/std/barrier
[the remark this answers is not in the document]
griwes: I'm not seeing any further changes in this file. Line 389, for one, isn't using nvtarget.

libcxx/include/support/atomic/atomic_base.h
```diff
@@ -0,0 +1,259 @@
...
+// under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef _LIBCUDACXX_ATOMIC_BASE_H
+#define _LIBCUDACXX_ATOMIC_BASE_H
+
+#include <type_traits>
```
griwes, line 14: If a file including this file doesn't include type_traits, type_traits is included inside a namespace, right? This needs to stop being here.

libcxx/include/support/atomic/atomic_cuda.h
```diff
@@ -0,0 +1,483 @@
...
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#if defined(__CUDA_MINIMUM_ARCH__) && ((!defined(_MSC_VER) && __CUDA_MINIMUM_ARCH__ < 600) || (defined(_MSC_VER) && __CUDA_MINIMUM_ARCH__ < 700))
+#  error "CUDA atomics are only supported for sm_60 and up on *nix and sm_70 and up on Windows."
+#endif
+
+#ifndef __CUDACC_RTC__
+#include <string.h>
```
griwes, line 15: Same comment as for type_traits above here.

include/cuda/std/barrier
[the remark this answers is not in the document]
wmaxey: Hmm... That's because barrier fixes are part of another patch set. I won't be pulling those ones into it. Instead I can reset this file to main to avoid complicating things further.

libcxx/include/support/atomic/atomic_base.h
wmaxey: :+1: Thanks for the catch.

griwes (approved):
LGTM, good job!

wmaxey:
https://builds4u.nvidia.com/dvs/#/change/3025952755833026.2?eventType=Virtual

Build is all clear.

PR #187, opened 2021-07-20
Several fixes for Clang 12 and GCC 11

wmaxey:
I haven't properly diagnosed the tuple CTAD failure in GCC 11, but since the rest of the suite passes I'm skipping this case for this compiler.

The nodebug type error will occur every release, I've disabled it from the `<cuda/std/...>` to avoid having to make the same bump every time.


wmaxey:
Internal CI passed.
https://scbuilds4u/dvs/#/change/3021210839432407.1?eventType=Virtual

griwes (approved):

PR #194, opened 2021-08-10
Fix atomic and barrier on NVRTC.

wmaxey:
Something in NVRTC has changed that requires a new method of querying if the barrier layout is messy.

Atomics are unable to select a proper backend when built with NVRTC. This is fixed with a stub include of what is in `atomic_base.h`.


libcxx/include/support/atomic/atomic_cuda.h
```diff
@@ -97,6 +97,8 @@ namespace __host {
 #elif defined (_LIBCUDACXX_HAS_C11_ATOMIC_IMP)
 //TODO
 // #  include "atomic_c11.h"
+#elif defined(_LIBCUDACXX_COMPILER_NVRTC)
+#  include "atomic_nvrtc.h"
```
griwes, line 101: Ooooh. It's because prior to the refactor, all the host bits were ifdef'd out, but now we depend on storage types and whatnot from `__host`? Makes sense.

griwes (approved):

libcxx/include/support/atomic/atomic_cuda.h
wmaxey: Right, we should make an abstraction for storage. I remember attempting to do that and finding it too difficult to remove the scope parameter in the same change.

PR #197, opened 2021-08-18
Fix `cuda::atomic<Integral>`'s `fetch_max/fetch_min` extension

wmaxey:
There was previously no testing, and this has been broken since 1.3.0. I am unable to test back further.

### DIFF
--- a/.upstream-tests/test/cuda/atomics/atomic.ext/atomic_fetch_max.pass.cpp
+++ b/.upstream-tests/test/cuda/atomics/atomic.ext/atomic_fetch_max.pass.cpp
@@ -1,0 +1,74 @@
+//===----------------------------------------------------------------------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+//
+// UNSUPPORTED: libcpp-has-no-threads, pre-sm-60
+// UNSUPPORTED: windows && pre-sm-70
+
+// <cuda/std/atomic>
+
+#include <cuda/std/atomic>
+#include <cuda/std/type_traits>
+#include <cuda/std/cassert>
+
+#include "test_macros.h"
+#include "atomic_helpers.h"
+#include "cuda_space_selector.h"
+
+template <class T, template<typename, typename> typename Selector, cuda::thread_scope>
+struct TestFn {
+  __host__ __device__
+  void operator()() const {
+    // Test greater
+    {
+        typedef cuda::atomic<T> A;
+        Selector<A, constructor_initializer> sel;
+        A & t = *sel.construct();
+        t = T(1);
+        assert(t.fetch_max(2) == T(1));
+        assert(t.load() == T(2));
+    }
+    {
+        typedef cuda::atomic<T> A;
+        Selector<volatile A, constructor_initializer> sel;
+        volatile A & t = *sel.construct();
+        t = T(1);
+        assert(t.fetch_max(2) == T(1));
+        assert(t.load() == T(2));
+    }
+    // Test not greater
+    {
+        typedef cuda::atomic<T> A;
+        Selector<A, constructor_initializer> sel;
+        A & t = *sel.construct();
+        t = T(3);
+        assert(t.fetch_max(2) == T(3));
+        assert(t.load() == T(3));
+    }
+    {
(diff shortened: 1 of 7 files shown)


.upstream-tests/test/cuda/atomics/atomic.ext/atomic_fetch_max.pass.cpp
```diff
@@ -0,0 +1,74 @@
...
+struct TestFn {
+  __host__ __device__
+  void operator()() const {
+    // Test greater
+    {
+        typedef cuda::atomic<T> A;
+        Selector<A, constructor_initializer> sel;
+        A & t = *sel.construct();
+        t = T(1);
+        assert(t.fetch_max(2) == T(1));
```
wmaxey, line 32: Is this expected? I believe the tests should operate similarly to `fetch_add` where the result is the current value of the atomic.

libcxx/include/support/atomic/atomic_cuda.h
```diff
@@ -396,7 +396,7 @@ __host__ __device__
     _Tp __expected = __cxx_atomic_load(__a, memory_order_relaxed);
     _Tp __desired = __expected < __val ? __expected : __val;
 
-    while(__desired != __val &&
+    while(__desired == __val &&
```
wmaxey, line 399: Either I've introduced a bug or there is something else wrong here. This was different from the implementation for max.

.upstream-tests/test/cuda/atomics/atomic.ext/atomic_fetch_max.pass.cpp
griwes: This looks right to me.

libcxx/include/support/atomic/atomic_cuda.h
```diff
@@ -376,6 +376,34 @@ __host__ __device__
     )
 }
 
+template <typename _Tp, typename _Delta, int _Sco, bool _Ref>
+__host__ __device__
+ _Tp __cxx_atomic_fetch_max(__cxx_atomic_base_heterogeneous_impl<_Tp, _Sco, _Ref> volatile* __a, _Delta __val, memory_order __order) {
+    _Tp __expected = __cxx_atomic_load(__a, memory_order_relaxed);
```
griwes, line 382: This needs to be scoped, no?

libcxx/include/support/atomic/atomic_cuda.h
```diff
@@ -376,6 +376,34 @@ __host__ __device__
...
+    _Tp __desired = __expected > __val ? __expected : __val;
+
+    while(__desired == __val &&
+            !__cxx_atomic_compare_exchange_strong(__a, &__expected, __desired, __order, __order)) {
```
griwes, line 386: Same here (and same in fetch_min).

libcxx/include/support/atomic/atomic_cuda.h
wmaxey: `_Sco` is carried by `__a` for the `__cxx_atomic_` layer.

.upstream-tests/test/cuda/atomics/atomic.ext/atomic_fetch_max.pass.cpp
wmaxey: If that's the case then the documentation and example is wrong.

https://nvidia.github.io/libcudacxx/extended_api/synchronization_primitives/atomic/fetch_max.html

libcxx/include/support/atomic/atomic_cuda.h
griwes: Ah! that makes a lot of sense.

.upstream-tests/test/cuda/atomics/atomic.ext/atomic_fetch_max.pass.cpp
griwes: The docs and the example suggest that this isn't an RMW operation, but then this wouldn't be very useful. The example is also clearly not how `atom.{min,max}` work in PTX.

griwes (approved):
Approve, we can fix the docs independently.

.upstream-tests/test/cuda/atomics/atomic.ext/atomic_fetch_max.pass.cpp
wmaxey: I hadn't dug deep enough on this. I bypassed the PTX enhancements. Will append another commit to fix.

griwes (approved):

PR #207, opened 2021-09-22
Fix GCC/Clang only compilation of <cuda/std/atomic>

wmaxey:


libcxx/include/support/atomic/atomic_cuda_scopes.h
```diff
@@ -0,0 +1,41 @@
+#ifndef __ATOMIC_BLOCK
```
griwes, line 1: I wonder if this name or something like `atomic_scopes.h` is better. We only expose these in the `cuda::` versions of the interface, true, but it's used throughout...

griwes (approved):
LGTM

PR #234, opened 2021-12-07
Fix host only compilation of <cuda/std/cmath>

wmaxey:
This patch set fixes compilation on clang and gcc. Currently there are missing/extra definitions caused by nvcc including math.h by default.

VDVS: https://builds4u.nvidia.com/dvs/#/change/3067999458696529.1?eventType=Virtual&dvs_showStaging=on

### DIFF
--- a/include/cuda/std/detail/libcxx/include/__config
+++ b/include/cuda/std/detail/libcxx/include/__config
@@ -792,6 +792,11 @@ typedef __char32_t char32_t;
 #define _LIBCUDACXX_HAS_NO_ASAN
 #endif
 
+// GCC implemented P0030R1 in 7.0, maybe only available under experimental?
+#if _GNUC_VER < 700
+#define _LIBCUDACXX_NO_HOST_CPP17_HYPOT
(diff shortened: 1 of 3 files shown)


include/cuda/std/detail/libcxx/include/__config
```diff
@@ -792,6 +792,11 @@ typedef __char32_t char32_t;
 #define _LIBCUDACXX_HAS_NO_ASAN
 #endif
 
+// GCC implemented P0030R1 in 7.0, maybe only available under experimental?
+#if _GNUC_VER < 700
+#define _LIBCUDACXX_NO_HOST_CPP17_HYPOT
```
griwes, line 797: I can see _a_ use of this in cmath that's being removed in this PR. Is this still needed?

griwes (approved):

include/cuda/std/detail/libcxx/include/__config
wmaxey: Removed!

PR #237, opened 2021-12-16
Backports to C++11 of <bit>

wmaxey:
# <cuda/std/bit> - C++20 backport of bit manipulation intrinsics

### Allows manipulation of unsigned values using [`<bit>`](https://en.cppreference.com/w/cpp/header/bit)

# Details

## Motivation

- CUDA has intrinsics for `popc/clz/ctz` but they differ from host compiler builtins. Libcu++ has an opportunity to unify host and device code for these intrinsics.
- `int128` in device code is not able to make use of CUDA intrinsics, this backport provides a general method for users to manipulate wider values.

## Impact

- Backports of <bit> to C++11 requires that most functions are minimized into one statement, this impacts readability of the header.
- Conditional use of `is_constant_evaluated` is required to improve performance where possible.
- Test coverage should remain the same, except for added coverage in C++11 and up.

### Testing

- [x] Bug fixes have regression tests that would reproduce the bug.
- [x] New features should have correctness tests to validate behavior.
- [ ] Benchmarks have been added to monitor performance of new features. (not required)


include/cuda/std/detail/libcxx/include/bit
```diff
@@ -53,58 +53,300 @@ namespace std {
...
+#  endif
+#else
+  return __builtin_ctzl(__x);
+#endif
+}
 
-_LIBCUDACXX_BEGIN_NAMESPACE_STD
+inline _LIBCUDACXX_INLINE_VISIBILITY _LIBCUDACXX_BIT_CONSTEXPR
+int __libcpp_ctz(unsigned long long __x) _NOEXCEPT {
+#if defined(__CUDA_ARCH__)
```
griwes, line 164: Should these be if target instead of checks for `__CUDA_ARCH__`?

griwes (approved):
I assume that the bit fiddling in here works and is verified by tests.

In addition to the inline comment - rename the tests in libcxx/test to match the names that actually ended up in the standard.

include/cuda/std/detail/libcxx/include/bit
wmaxey: This is behavior specific to NVCC. (Also NV_IF_TARGET embeds brackets due to if target requiring them `{...}` and this breaks C++11 constexpr.

PR #241, opened 2022-01-18
Use decimal notation for version number.

allisonvacanti:
```
error: invalid digit "8" in octal constant
  001008000
  ^~~~~~~~~
```

PR #242, opened 2022-01-18
Add `find_package` and `add_subdirectory` CMake support.

allisonvacanti:
To validate installation + packages, run the new tests:

```
mkdir build
cd build
cmake -Dlibcudacxx_ENABLE_CMAKE_TESTS=ON ..
ctest --output-on-failure
```

Tests are failing until #241 is merged.


cmake/libcudacxxInstallRules.cmake
```diff
@@ -0,0 +1,38 @@
...
+
+if (NOT libcudacxx_ENABLE_INSTALL_RULES)
+  return()
+endif()
+
+# Bring in CMAKE_INSTALL_LIBDIR
+include(GNUInstallDirs)
+
+# libcudacxx is a header library; no need to build anything before installing:
+set(CMAKE_SKIP_INSTALL_ALL_DEPENDENCY TRUE)
```
robertmaynard, line 13: This logic seems to be tied to the fact that `libcudacxx` isn't the top project and not the fact we are asking for install rules. We need to either guard this with `TOPLEVEL_PROJECT` or move it to where we set `TOPLEVEL_PROJECT`

cmake/test/test_export/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
+cmake_minimum_required(VERSION 3.20.1)
+project(libcudacxxTestExport CXX)
+
+message(STATUS "libcudacxx_DIR=${libcudacxx_DIR}")
+find_package(libcudacxx)
```
robertmaynard, line 5: Any reason to not mark this as `REQUIRED`?

cmake/libcudacxxInstallRules.cmake
allisonvacanti: My understanding is that this flag just disables the dependency on building `all` when doing `make install`, and since we're just setting a local variable, this effect should be limited to the libcudacxx install rules (AIUI).

The install rules are disabled by default when `TOPLEVEL` is false, so this shouldn't impact any users unless they opt-in. I want to have this available to subprojects since it might be useful for generating temporary Thrust installs during testing.

Is there an issue I'm missing?

cmake/test/test_export/CMakeLists.txt
allisonvacanti: Good catch, it should be. I'll update this.

cmake/libcudacxxInstallRules.cmake
robertmaynard: Since this is inside an `include()` at the root of the project this variable is now defined for all subdirectories of the repository. Which means that when building this is the top/root project this value will be set for `add_subdirectory(libcxx)` and the subsequent creation of the `libcxx` and `libcxx_static` libraries.

allisonvacanti: That's a good point, makes sense. I'll just take this out for now, since it's just a "nice to have." If we can find a way to set this per-project I'll add it back.

allisonvacanti:
Last push fixed a bug in the versions checks (cc: @robertmaynard)

wmaxey:
VDVS: http://scdvs.nvidia.com/Build_Results?virtualId=1000081020&which_page=current_build
VDVS: http://ausdvs.nvidia.com/Build_Results?virtualId=1001080494&which_page=current_build

lib/cmake/libcudacxx/libcudacxx-config.cmake
```diff
@@ -0,0 +1,56 @@
...
+  #    system includes. This means that the Toolkit libcudacxx will *always* be used
+  #    during compilation, and the include paths of an IMPORTED libcudacxx::libcudacxx
+  #    target will never have any effect.
+  # 4) This behavior can be fixed by setting the property NO_SYSTEM_FROM_IMPORTED
+  #    on EVERY target that links to libcudacxx::libcudacxx. This would be a burden and a
+  #    footgun for our users. Forgetting this would silently pull in the wrong libcudacxx!
+  # 5) A workaround is to make a non-IMPORTED library outside of the namespace,
+  #    configure it, and then ALIAS it into the namespace (or ALIAS and then
+  #    configure, that seems to work too).
+  add_library(${ugly_name} INTERFACE)
```
robertmaynard, line 25: This approach won't work when the consumer is part of an install or build export set.
Adding the following to the `test_export` exposes the issue:

```cmake
# Install tree exporting of version_check_objs
install(TARGETS version_check_objs EXPORT test-targets)
install(EXPORT test-targets FILE test-targets.cmake DESTINATION "lib/cmake/libcudacxx-test/")

# Build tree exporting of version_check_objs
export(EXPORT test-targets FILE "${CMAKE_CURRENT_BINARY_DIR}/libcudacxx-test/test-targets.cmake")
```

The reason this approach works for Thrust is that every project has to construct a <namespace>::Thrust target 
via `thrust_create_target` that is IMPORTED INTERFACE. I believe that what we can do is just make ` alias_name` the firewall and things should work for the above use-cases.

```cmake
  add_library(${ugly_name} INTERFACE)

  add_library(${alias_name} INTERFACE IMPORTED GLOBAL)
  target_link_libraries(${alias_name} INTERFACE  ${ugly_name})
```

robertmaynard (changes requested):

lib/cmake/libcudacxx/libcudacxx-config.cmake
allisonvacanti: Makes sense -- fixed in last push.

CMakeLists.txt
```diff
@@ -1,6 +1,33 @@
...
 
+# Determine whether libcudacxx is the top-level project or included into
+# another project via add_subdirectory().
+if ("${CMAKE_SOURCE_DIR}" STREQUAL "${CMAKE_CURRENT_LIST_DIR}")
+  set(libcudacxx_TOPLEVEL_PROJECT ON)
+else()
+  set(libcudacxx_TOPLEVEL_PROJECT OFF)
+endif()
+
+include(cmake/libcudacxxInstallRules.cmake)
```
robertmaynard, line 12: We need to ensure that this file `lib/cmake/libcudacxx.cmake` is invoked each time libcudacxx is consumed via `add_subdirectory`

Consider when a project uses libcudacxx via CPM.

```
set(libcudacxx_ENABLE_INSTALL_RULES ON)
find_package(libcudacxx) #fails 
# cpm git clones libcudacxx
add_subdirectory(libcudacxx)
include(cmake/libcudacxxInstallRules.cmake) 
include(cmake/libcudacxxAddSubdir.cmake)
set(libcudacxx_DIR <path to libcudacxx-config.cmake> CACHE)
```

Now on first configuration this means we have install rules for libcudacxx. But if we reconfigure the project
this what happens:

```
set(libcudacxx_ENABLE_INSTALL_RULES ON)
find_package(libcudacxx) # works since libcudacxx_DIR is a cache var
```
Now we have no install rules when we do `cmake . && cmake . && cmake --install .` compared to we had install rules when we did `cmake . && cmake --install .`

Maybe after  `include(cmake/libcudacxxAddSubdir.cmake)` we unset the cache variable. We need to see if that works.

robertmaynard (changes requested):

CMakeLists.txt
allisonvacanti: We can just remove this when for non-toplevel usages to unblock. I only wanted this for testing Thrust installations, but we can find a better solution later when I start testing that. I don't think anyone else will want to add libcudacxx install rules to their builds.

allisonvacanti: The last push moved the install rules to after the non-toplevel early return.

robertmaynard: When libcudacxx is consumed via `add_subdirectory` not having install rules means that the consumer could depend on a `libcudacxx::libcudacxx` target that doesn't exist. So when that project installs itself, it will have broken dependencies since `libcudacxx` hasn't been installed ( either the headers or the `libcudacxx-config.cmake` )

allisonvacanti: I thought that this would be taken care of by the consumers export config. Let's discuss more when we chat later today.

allisonvacanti: c7d9ddd implements the changes @robertmaynard and I discussed last week:

- Re-enable install rules for `add_subdirectory` builds
- Switch from `find_package` to explicitly including the `libcudacxx-config*` files, bypassing the `find_package` machinery that sets `libcudacxx_DIR` and breaks CPM.

robertmaynard (approved):

allisonvacanti:
@wmaxey This should be ready to go -- I've been using the package in thrust successfully in https://github.com/NVIDIA/thrust/pull/1605 👍

allisonvacanti:
BTW @wmaxey -- you'll need to ping the builds folks to make sure that these three files get packaged into CTK. See NVBug 3273504 where we did this for Thrust/CUB.

```
$SRC_DIR/lib/cmake/libcudacxx/libcudacxx-config.cmake         -> $CTK_ROOT/lib/cmake/libcudacxx/libcudacxx-config.cmake
$SRC_DIR/lib/cmake/libcudacxx/libcudacxx-config-version.cmake -> $CTK_ROOT/lib/cmake/libcudacxx/libcudacxx-config-version.cmake
$SRC_DIR/lib/cmake/libcudacxx/libcudacxx-header-search.cmake  -> $CTK_ROOT/lib/cmake/libcudacxx/libcudacxx-header-search.cmake
```

PR #246, opened 2022-02-09
Two small wording changes

daniellustig:
1) Clarify that atomics in any GPU's memory (not just in peer GPU
memory) are also atomic if only accessed by GPU threads

2) Remove an extra period


wmaxey:
Thanks Daniel.

wmaxey (approved):

PR #252, opened 2022-02-18
Fix deprecation warnings in C++20

wmaxey:
Adds a config for C++20 under clang-13.

There are several deprecation fixes as well to ensure libcxx builds clean.


griwes (approved):

libcxx/test/std/utilities/function.objects/func.search/func.search.bm/default.pass.cpp
```diff
@@ -7,7 +7,7 @@
 //===----------------------------------------------------------------------===//
 
 // UNSUPPORTED: c++98, c++03, c++11, c++14
-// XFAIL: c++17, c++2a
+// XFAIL: c++17, c++2a, c++20
```
brycelelbach, line 10: Do we need to leave `c++2a` in here?

brycelelbach (commented):
LGTM at a glance.

Issue #24, opened 2020-09-25
Add a light mode to libcu++ website and make it more friendly to color vision deficiency


harrism:
A light mode would be nice. 

I wonder if the libcu++ site is friendly to people with color vision deficiency?  (I'm always wary of NVIDIA green, especially on top of black or dark grey, for this reason.)  Even I find the green on that site a bit hard to read, and I don't think I have CVD. 

This link shows the site filtered to demonstrate various forms of CVD. https://www.toptal.com/designers/colorfilter?orig_uri=https://nvidia.github.io/libcudacxx/api.html&process_type=grey

PR #258, opened 2022-03-20
Increase contrast of search input text in docs.

bdice:
This PR increases the contrast of the documentation's search bar text so that the input text is legible.

If this PR is approved/merged, I will make a similar PR to improve Thrust's documentation. If anyone is aware of other instances of this issue, let me know and I can make corresponding updates to other repositories.

Before:
image

After:
image

### DIFF
--- a/docs/_sass/color_schemes/nvidia.scss
+++ b/docs/_sass/color_schemes/nvidia.scss
@@ -121,3 +121,4 @@ pre.highlight code
 .highlight span.gd { color: #ff0000 } /* Generic.Deleted */
 .highlight span.gi { color: #00ff00 } /* Generic.Inserted */
 
+.search-input { color: $body-text-color; }


wmaxey:
I'll take a look and see if this is browser unique, I don't seem to encounter this on Chrome. In fact I have the opposite problem, the search pre-fill is legible while my query is black on grey and quite difficult to read.

bdice:
Interesting observation, the default input colors may differ by browser (I saw this in Firefox). Setting a color in the CSS should hopefully fix the problem for all browsers. 😊

harrism:
I definitely see the unreadable black on charcoal search box in Firefox. IMO we should get rid of the default black background website because it leads to usability issues. See also #24 .

bdice:
Following up from the question about Chrome: I verified that this fix works for Firefox, Chrome, and Edge browsers. All three browsers show light text on a dark background in the placeholder text and when input is entered. Here's what it looks like when input is entered into the search box:
image

wmaxey:
Thank you both for the feedback, I agree that the docs need a readability update, but at the moment that's somewhat beyond me. I'll take some time to find out if I can do it myself or bug Bryce about it.

wmaxey (approved):

PR #259, opened 2022-03-22
Forcefully disable warnings in GCC for <nv/target>

wmaxey:
Currently GCC 6 and down produces this warning when compile with `-Wpedantic`

```
<source>:594:1: warning: ISO C++11 requires at least one argument for the "..." in a variadic macro
     )
```

While it's possible to ignore `-Wpedantic`, I believe it would be more reliable to silence warnings by marking the details of <nv/target> as `system_header`s.

### DIFF
--- a/include/nv/detail/__preprocessor
+++ b/include/nv/detail/__preprocessor
@@ -7,6 +7,9 @@
 //
 //===----------------------------------------------------------------------===//
 
+#if defined(__GNUC__) && !defined(__clang__)
+#pragma GCC system_header
+#endif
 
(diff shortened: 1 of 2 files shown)


griwes (approved):

dkolsen-pgi (approved):
Clang seems to recognize and honor `#pragma GCC system_header`.  (At least Clang 12 on Linux does.)  Is the `!defined(__clang__)` necessary?

wmaxey:
I would be willing to remove the condition, but it doesn't seem necessary as I haven't encountered warnings on Clang over this.

PR #267, opened 2022-04-15
Docker refactor, parameterizes OS and compiler versions.

wmaxey:
Ideally we instead generate one layer that handles all dialects.

CMake demands a CUDA Toolkit be present, currently internal builds of the runtime do not satisfy that requirement so we must still sync from nvcr.io.


docker2/config.Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+FROM cccl/base AS libcudacxx-configured
+
+ARG HOST_CXX
+ARG CXX_DIALECT
+# Attempt to load env from cccl/cuda
+ARG CUDACXX_PATH
+
+ADD . /libcudacxx
+
+# Install compiler and configure project
```
miscco, line 28: Does this really install the compiler?

docker2/driver.Dockerfile
```diff
@@ -0,0 +1,12 @@
+# Copyright (c) 2018-2020 NVIDIA Corporation
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+# Released under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+
+# ARG BASE_IMAGE=nvcr.io/nvidia/nvhpc:22.3-devel-cuda_multi-ubuntu20.04
```
miscco, line 6: Should that be something conditional like below
```
if [ ${NVCXX_ENABLED} ]; then
BASE_IMAGE="nvcr.io/nvidia/nvhpc:22.3-devel-cuda_multi-ubuntu20.04"
else
BASE_IMAGE="nvcr.io/nvidia/cuda:11.7.0-devel-ubuntu22.04"
fi;
```

docker2/wrapper.sh
```diff
@@ -0,0 +1,128 @@
...
+################################################################################
+# Begin building container
+
+section_separator
+
+TEMP_DIR=$(mktemp -d)
+EMPTY_DIR=$(mktemp -d)
+
+LIBCUDA=$(ldconfig -p | grep libcuda.so | tr ' ' '\n' | grep / | head -n 1 | tr '\n' ' ' | sed 's/ *$//')
+LIBNVIDIAPTXJITCOMPILER=$(ldconfig -p | grep libnvidia-ptxjitcompiler.so | tr ' ' '\n' | grep / | head -n 1 | tr '\n' ' ' | sed 's/ *$//')
```
miscco, line 115: Could we add some underscores here to make it somewhat readable?

miscco (commented):
This already looks great.

I am super looking forward to using this

docker2/driver.Dockerfile
wmaxey: There's no conditionals in dockerfiles, it can only be parameterized. 

At least I think.

wmaxey:
@miscco Give this a shot, on Windows + VSCode it's pretty stellar.

environments/linux/docker/.dockerignore
```diff
@@ -0,0 +1,2 @@
+.git
+.vscode/*
```
miscco, line 2: Should we add trailing newlines to all files?

miscco (approved):
I am loving it!

environments/linux/docker/.dockerignore
wmaxey: Fixed in last commit.

PR #272, opened 2022-04-27
Fix device code path in <nv/target> when compiling with clang-cuda

wmaxey:
Fixes https://nvbugswb.nvidia.com/NvBugs5/SWBug.aspx?bugid=3611649&cmtNo=

### DIFF
--- a/include/nv/target
+++ b/include/nv/target
@@ -18,6 +18,9 @@
 #  define _NV_COMPILER_NVCC
 #elif defined(__NVCOMPILER) && __cplusplus >= 201103L
 #  define _NV_COMPILER_NVCXX
+#elif defined(__clang__) && defined(__CUDA__) && defined(__CUDA_ARCH__)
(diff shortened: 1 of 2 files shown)


include/nv/target
```diff
@@ -18,6 +18,11 @@
 #  define _NV_COMPILER_NVCC
 #elif defined(__NVCOMPILER) && __cplusplus >= 201103L
 #  define _NV_COMPILER_NVCXX
+#elif defined(__clang__) && defined(__CUDA__) && defined(__CUDA_ARCH__)
+// clang compiling CUDA code, device mode.
+#  define _NV_COMPILER_CLANG_CUDA
+// Also define NVCC to enable the same code paths.
```
griwes, line 24: If we did this the other possible way and changed all the places we enter the nvcc code path to also check for clang, how big would the diff be?

griwes (approved):
Squash it and ship it!

PR #299, opened 2022-08-09
Add `mdspan` reference implementation

youyu3:
* Pulls the mdspan reference implementation from branch "stable" of the kokkos repo, https://github.com/kokkos/mdspan, up to PR 172.
* Uglified internal identifiers and made some naming convention updates.


include/cuda/std/detail/libcxx/include/experimental/__p0009_bits/config.hpp
```diff
@@ -0,0 +1,302 @@
...
+// GCC 10's CTAD seems sufficiently broken to prevent its use.
+#  if (defined(__MDSPAN_COMPILER_CLANG) || !defined(__GNUC__) || __GNUC__ >= 11) \
+      && ((defined(__cpp_deduction_guides) && __cpp_deduction_guides >= 201703) \
+         || (!defined(__cpp_deduction_guides) && __MDSPAN_HAS_CXX_17))
+#    define __MDSPAN_USE_CLASS_TEMPLATE_ARGUMENT_DEDUCTION 1
+#  endif
+#endif
+
+#ifndef __MDSPAN_USE_ALIAS_TEMPLATE_ARGUMENT_DEDUCTION
+// GCC 10's CTAD seems sufficiently broken to prevent its use.
```
mhoemmen, line 228: @youyu3 We may want to excise some of the original author's snark for public releases.

include/cuda/std/experimental/mdspan
```diff
@@ -0,0 +1,56 @@
...
+// NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+// SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+//
+// Questions? Contact Christian R. Trott (crtrott@sandia.gov)
+//
+// ************************************************************************
+//@HEADER
+*/
+
+#pragma once
```
wmaxey, line 44: Probably fine, but we tend to use `#ifndef/def` guards.

include/cuda/std/experimental/mdspan
```diff
@@ -0,0 +1,56 @@
...
+
```
wmaxey, line 45: There should be a `#include "[cuda/std/]detail/__config" here before anything else. This way libcudacxx macros get injected.

It might need to be `#include "../detail/__config" in this case.

include/cuda/std/detail/libcxx/include/experimental/__p0009_bits/default_accessor.hpp
```diff
@@ -0,0 +1,86 @@
...
+//@HEADER
+*/
+
+#pragma once
+
+#include "macros.hpp"
+
+#include <cstddef> // size_t
+
+namespace std {
```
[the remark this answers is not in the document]
wmaxey: This would be true for other headers that open namespaces.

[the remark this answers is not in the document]
youyu3: Since this is `namespace std { namespace experimental { } }`, should I be using `_LIBCUDACXX_BEGIN_NAMESPACE_EXPERIMENTAL` instead?

include/cuda/std/experimental/mdspan
youyu3: Added.

include/cuda/std/experimental/mdspan
youyu3: Updated to use `#ifndef/def`

include/cuda/std/detail/libcxx/include/experimental/__p0009_bits/default_accessor.hpp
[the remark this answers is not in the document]
youyu3: Updated.

include/cuda/std/detail/libcxx/include/experimental/__p0009_bits/extents.hpp
```diff
@@ -90,7 +91,7 @@ static std::integral_constant<
   )
 >
 _check_compatible_extents(
-  std::true_type, std::integer_sequence<size_t, _Extents...>, std::integer_sequence<size_t, _OtherExtents...>
+  std::true_type, ::std::integer_sequence<size_t, _Extents...>, ::std::integer_sequence<size_t, _OtherExtents...>
```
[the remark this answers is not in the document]
youyu3: Makes sense.  Thanks.  I'll submit a change with all occurrences of `true_type`, `false_type`, and `size_t` updated.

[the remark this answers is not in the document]
mhoemmen: @youyu3 Thanks! : - )

include/cuda/std/detail/libcxx/include/experimental/__p0009_bits/extents.hpp
```diff
@@ -247,7 +248,7 @@ class extents
     /* requires */ (
       // TODO: check whether the other version works with newest NVCC, doesn't with 11.4
       // NVCC seems to pick up rank_dynamic from the wrong extents type???
-      __MDSPAN_FOLD_AND(__MDSPAN_TRAIT(is_convertible, _Integral, index_type) /* && ... */) &&
+      __MDSPAN_FOLD_AND(__MDSPAN_TRAIT(::std::is_convertible, _Integral, index_type) /* && ... */) &&
       __MDSPAN_FOLD_AND(__MDSPAN_TRAIT(is_nothrow_constructible, index_type, _Integral) /* && ... */) &&
```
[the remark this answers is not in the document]
youyu3: Fixed.

include/cuda/std/experimental/mdspan
```diff
@@ -1,57 +1,16 @@
...
-//@HEADER
-*/
+//===----------------------------------------------------------------------===//
 
-#pragma once
+#ifndef _CUDA_EXPERIMENTAL_MDSPAN
+#define _CUDA_EXPERIMENTAL_MDSPAN
 
 #include "../detail/__config"
-#include "../detail/libcxx/include/experimental/__p0009_bits/default_accessor.hpp"
```
wmaxey: Revert this, when accessing mdspan through this header, we ensure all dependencies for `mdspan` are handled here first.

include/cuda/std/detail/libcxx/include/experimental/mdspan
```diff
@@ -0,0 +1,56 @@
...
+//
+// Questions? Contact Christian R. Trott (crtrott@sandia.gov)
+//
+// ************************************************************************
+//@HEADER
+*/
+
+#pragma once
+
+#include "__p0009_bits/default_accessor.hpp"
```
wmaxey, line 46: This needs to be wrapped with:

`#ifndef __cuda_std__`

include/cuda/std/detail/libcxx/include/experimental/__p0009_bits/trait_backports.hpp
```diff
@@ -0,0 +1,160 @@
...
+//@HEADER
+*/
+
+#ifndef _LIBCUDACXX_EXPERIMENTAL___P0009_BITS_TRAIT_BACKPORTS_HPP
+#define _LIBCUDACXX_EXPERIMENTAL___P0009_BITS_TRAIT_BACKPORTS_HPP
+
+#include "macros.hpp"
+#include "config.hpp"
+
+#include <type_traits>
```
wmaxey, line 50: Similarly, wrap with `#ifndef __cuda_std__` and add includes for these in `libcudacxx/include/cuda/std/experimental/mdspan`

include/cuda/std/detail/libcxx/include/experimental/mdspan
youyu3: Does this refer to all header being included here?

So it's like we have
```
#ifndef __cuda_std__
#include "__p0009_bits/default_accessor.hpp"
#include "__p0009_bits/other.hpp"
...
#endif
```
here, and have (according to your other comment)
```
#include "../detail/libcxx/include/experimental/__p0009_bits/default_accessor.hpp"
#include "../detail/libcxx/include/experimental/__p0009_bits/other.hpp"
...
```
in `include/cuda/std/experimental/mdspan`?

In that case, I don't think the other `mdspan` header needs to include this file, since they both include the same set of headers?

include/cuda/std/detail/libcxx/include/experimental/__p0009_bits/trait_backports.hpp
youyu3: Is this because different `type_traits` gets included when included from here vs from `include/cuda/std/experimental/mdspan`?
Like one picks up `include/cuda/std/detail/libcxx/include/type_traits` and the other picks up `include/cuda/std/type_traits?

include/cuda/std/detail/libcxx/include/experimental/__p0009_bits/extents.hpp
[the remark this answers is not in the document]
mhoemmen: Awesome, thanks! : - )

include/cuda/std/detail/libcxx/include/experimental/__p0009_bits/trait_backports.hpp
youyu3: From what I see in other files, I think I should
- In `libcudacxx/include/cuda/std/experimental/mdspan`, have `#include "type_traits"`
- In files under `__p0009_bits`, have `#include <type_traits>` when `#ifndef __cuda_std__`.

.upstream-tests/test/std/containers/views/mdspan/mdspan.ctor/ctad_extents.pass.cpp
```diff
@@ -0,0 +1,64 @@
...
+    assert(m.rank_dynamic() == 2       ); \
+    assert(m.extent(0)      == 64      ); \
+    assert(m.extent(1)      == 128     )
+
+namespace stdex = cuda::std::experimental;
+
+int main(int, char**)
+{
+    // workaround to avoid compilation errors
+    cuda::std::experimental::detail::__make_dynamic_extent<int>();
```
mhoemmen, line 25: @youyu3 Could you please add details of the compiler version and flags here?  This feels very much like a compiler bug.  `__make_dynamic_extent<T>()` literally just returns `T(-1)`.  Its definition should be available, so it shouldn't need to be explicitly instantiated.

mhoemmen: @youyu3 If `__make_dynamic_extent` doesn't work, we could try this:

```c++
template<class... Integrals>
  requires((is_­convertible_­v<Integrals, size_­t> && ...))
  explicit extents(Integrals...) -> extents<IndexType, std::size_t(indices, -1)...>;
```

What matters here is on the right side of the `->`.  The (C++17) fold expression `size_t(indices, -1)...` expands to something like `size_t(-1)`, `size_t(-1)`, ..., `size_t(-1)`, where the number of `-1`s is `sizeof...(indices)`.

mhoemmen: Revision:
```c++
template <class... IndexTypes>
extents(IndexTypes...)
 -> extents<std::size_t, std::size_t((IndexTypes(), -1))...>;
```
The comma expression needed an extra (), else it was being interpreted as a cast with two arguments (not a valid cast).

youyu3: Thanks!  Pushed the workaround.

include/cuda/std/detail/libcxx/include/experimental/__p0009_bits/standard_layout_static_array.hpp
```diff
@@ -53,12 +53,16 @@
 #  include "no_unique_address.hpp"
 #endif
 
-#include <cuda/std/array>
+#ifndef __cuda_std__
+#include <array>
+#endif
 #ifdef __cpp_lib_span
 #include <span>
```
mhoemmen, line 60: Do we also need to protect `#include <span>` with `#ifndef __cuda_std__` ... `#endif` ?

youyu3: We don't have a `libcudacxx` version of `span` (or I didn't find one).

mhoemmen: @youyu3 Oh, oof : - (  Should we file an issue to add one?  We would need it for mdspan just for a few things that take `span` as input.

.upstream-tests/test/std/containers/views/mdspan/mdspan.layout.right.ctor/list_init.pass.cpp
```diff
@@ -0,0 +1,34 @@
...
+{
+    {
+        typedef int    data_t ;
+        typedef size_t index_t;
+
+        stdex::layout_right::mapping<stdex::extents<size_t,dyn, dyn>> m{stdex::dextents<size_t,2>{16, 32}};
+
+        static_assert( m.is_exhaustive() == true );
+
+        assert( m.extents().rank()         == 2  );
```
mhoemmen, line 25: [OPTIONAL] `rank()` or `rank_dynamic()` are always `static constexpr` and thus can be checked with `static_assert`.

.upstream-tests/test/std/containers/views/mdspan/mdspan.layout.right.ctor/list_init.pass.cpp
```diff
@@ -0,0 +1,34 @@
...
+
+int main(int, char**)
+{
+    {
+        typedef int    data_t ;
+        typedef size_t index_t;
+
+        stdex::layout_right::mapping<stdex::extents<size_t,dyn, dyn>> m{stdex::dextents<size_t,2>{16, 32}};
+
+        static_assert( m.is_exhaustive() == true );
```
mhoemmen, line 23: `is_exhaustive()` happens to be `static constexpr` for `layout_right`, but please note that this is not generally true for all layout mappings.

include/cuda/std/detail/libcxx/include/experimental/__p0009_bits/layout_left.hpp
```diff
@@ -73,21 +73,21 @@ class layout_left::mapping {
     struct __rank_count {};
 
     template <::std::size_t _r, ::std::size_t _Rank, class _Ip, class... _Indices>
-    __MDSPAN_INLINE_FUNCTION
+    __MDSPAN_HOST_DEVICE
```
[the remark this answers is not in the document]
youyu3: That's right.  But I'm only adding `__host__ __device__` while matching the declaration before the change, where the function wasn't declared as `inline`.

.upstream-tests/test/std/containers/views/mdspan/mdspan.layout.right.ctor/list_init.pass.cpp
youyu3: Thanks.  I'll change this to `assert` then.

mhoemmen: It's OK here -- I'm just pointing this out in case you want to write generic tests.  Thanks!

miscco:
Generally speaking I am extremely concerned about the implementation complexity here. 

I only looked at `extends` in detail, because I did implement it [previously](https://godbolt.org/z/T9T4db864). 

Especially compile times are going to be incredibly difficult here.

Also what I am generally wondering is why we are putting this in namespace stdex. It was merged into the standard after all

mhoemmen:
Hi @miscco !  We talked a bit about `extents` offline.  An excellent first step would be to run the benchmarks both with the current version, and with the current version with your `extents` implementation substituted in.  It would also be interesting to see if your implementation performs well at lower optimization levels.  (This would cover the important use case of debug builds not being horribly slow, which is important for game development.)

> Also what I am generally wondering is why we are putting this in namespace stdex. It was merged into the standard after all

This is a user library (not a vendor's implementation of the Standard Library).  Thus, it's not allowed to put things in namespace `std`.

mdspan was merged into C++23.  The reference mdspan implementation works with C++14, 17, and 20 as well.  For those back-porting use cases, it's important that we don't break applications' builds, in case those applications adopt a later compiler version that provides an mdspan implementation.  Thus, we don't want to collide with `std::mdspan`.

mhoemmen:
@miscco I wonder if some of the complexity of the reference mdspan implementation comes from an earlier requirement to back-port to C++11 (with its much more restrictive requirements for `constexpr` functions).  That requirement has since been dropped.  The reference mdspan implementation currently requires at least C++14.

crtrott:
I think the comment about extents could be simpler are fair. Part of the problem was trying to work around a number of issues with empty base optimization in MSVC and Intel compiler. MSVC did get much better though in the last couple years. That said: looking at some assembly it does look like the reference implementation compiles cleaner:

https://godbolt.org/z/Gvsfrn31o

crtrott:
Thinking more about this: there was a lot of work on extents, to essentially make the .extent(i) function optimize really well. Without the complication I think you end up with something which goes through some loop structures to figure out which element to access. I.e. you kinda rebuild the mapping for index to "which element from the dynamic extents array do I need to access" every time or so? But am not 100% sure.

miscco:
Yeah I believe my `_Dynamic_index` function is the issue

crtrott:
Note: with -O3 and GCC 11 it does compile into the same instruction sequence. I just told Mark, I am totally open to entertaining a whole sale replacement on our side. I think we should underpin this with some data down the line, for example we are working on using mdspan in Kokkos which then would make a whole lot of applications use it. If we collect data from that side regarding compile time/binary size/performance and from your applications we should have a good basis to make decisions.

jrhemstad:
Can we defer any significant refactoring discussion to future work/PR? I don't want to delay @youyu3's PR any more than necessary.

mhoemmen:
@jrhemstad Don't worry!  These investigations should not affect the current release.  Discussions continue on the reference mdspan implementation issue https://github.com/kokkos/mdspan/issues/190 .

include/cuda/std/detail/libcxx/include/experimental/mdspan
```diff
@@ -0,0 +1,59 @@
...
+// the U.S. Government retains certain rights in this software.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are
+// met:
+//
+// 1. Redistributions of source code must retain the above copyright
+// notice, this list of conditions and the following disclaimer.
+//
+// 2. Redistributions in binary form must reproduce the above copyright
```
miscco, line 18: Just to be sure, is that license compatible with the usual one. 

Do we have SWIPAT clearance for adding. Binary distribution clauses scare me a lot

.upstream-tests/test/std/containers/views/mdspan/mdspan.layout.right.ctor/list_init.pass.cpp
youyu3: Switched to `static_assert`.

include/cuda/std/detail/libcxx/include/experimental/__p0009_bits/standard_layout_static_array.hpp
youyu3: Now we have and it worked for my initial tests.  More tests to be added.

libcxx/test/std/containers/views/mdspan/mdspan.extents.cons/convertible_to_size_t.pass.cpp
```diff
@@ -0,0 +1,48 @@
...
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#include <experimental/mdspan>
+#include <cassert>
+#include <array>
+#include <span>
```
mhoemmen, line 12:
```suggestion
#ifdef __cpp_lib_span
#include <span>
#endif
```
I don't think you can even include `<span>` unless that feature test macro is defined.

mhoemmen (changes requested):

libcxx/test/std/containers/views/mdspan/mdspan.extents.cons/convertible_to_size_t.pass.cpp
youyu3: Thanks Mark, you are right.  However, I don't think an inclusion of `<span>` is even needed here, since it's already included by `<mdspan>`.  Tests ran fine without this line.
Actually, `#include <array>` is not needed here either.

jrhemstad:
Hey @youyu3 sorry for the last minute change, but based on some internal conversation, I think we can drop the `experimental` namespace for `mdspan`.

youyu3:
> 

Okay.  Will do.  Then I suppose that  we should move the files out of the `experimental` sub-directory as well?

include/cuda/std/detail/libcxx/include/experimental/__p0009_bits/dynamic_extent.hpp
```diff
@@ -52,9 +52,6 @@
 #endif
 
 _LIBCUDACXX_BEGIN_NAMESPACE_STD
-namespace experimental {
-
-__MDSPAN_INLINE_VARIABLE constexpr auto dynamic_extent = ::std::numeric_limits<::std::size_t>::max();
```
youyu3: This is needed when `span` is not available.  Fixed in a later commit...

libcxx/test/std/containers/views/mdspan/mdspan.extents.cons/convertible_to_size_t.pass.cpp
youyu3: Yeah I'm aware of that.  And any test that uses `span` should be guarded by `#ifdef __cpp_lib_span` as well (which has been fixed).

include/cuda/std/experimental/mdspan
youyu3: Done.  Including `detail/libcxx/include/mdspan` here.

include/cuda/std/detail/libcxx/include/experimental/mdspan
youyu3: I'm having `include/cuda/std/mdspan` getting all lower-level headers through this file, so I think `#ifndef __cuda_std__` is needed here.

include/cuda/std/detail/libcxx/include/experimental/__p0009_bits/extents.hpp
```diff
@@ -75,7 +76,7 @@ struct _count_dynamic_extents<> {
 
 template <size_t... _Extents, size_t... _OtherExtents>
 static constexpr std::false_type _check_compatible_extents(
-  std::false_type, std::integer_sequence<size_t, _Extents...>, std::integer_sequence<size_t, _OtherExtents...>
+  std::false_type, ::std::integer_sequence<size_t, _Extents...>, ::std::integer_sequence<size_t, _OtherExtents...>
```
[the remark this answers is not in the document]
youyu3: Thanks and done.  I'm going to resolve this one for now and upstream these changes to the reference impl later on.

include/cuda/std/detail/libcxx/include/experimental/__p0009_bits/config.hpp
mhoemmen: To clarify: It's not a huge deal here (the comment is not too snarky) but we could change it to something like "GCC 10 is known not to work with CTAD for this case."

youyu3: Done.

mhoemmen (approved):

mhoemmen:
It's far too much for me to review in detail, but I'm really impressed and happy with all the new unit tests!

include/cuda/std/detail/libcxx/include/__mdspan/config.hpp
```diff
@@ -0,0 +1,297 @@
...
+#endif
+#endif
+
+#ifdef _MSVC_LANG
+#define __MDSPAN_CPLUSPLUS _MSVC_LANG
+#else
+#define __MDSPAN_CPLUSPLUS __cplusplus
+#endif
+
+#define __MDSPAN_CXX_STD_14 201402L
```
miscco, line 66: We should really use `_LIBCUDACXX_STD_VER` for that

include/cuda/std/mdspan
```diff
@@ -0,0 +1,27 @@
...
+#include "span"
+#include "cstddef"
+#include "limits"
+#include "tuple"
+#include "type_traits"
+#include "utility"
+#include "version"
+
+#include "detail/__config"
+#include "detail/libcxx/include/algorithm"
```
miscco, line 23: I I believe we are missing the `include "detail/__pragma_push"`

include/cuda/std/detail/libcxx/include/type_traits
```diff
@@ -4767,6 +4767,8 @@ struct __has_operator_addressof
 
 #endif  // _LIBCUDACXX_CXX03_LANG
 
+template <class...> using _Void_t = void;
```
miscco, line 4770: I believe this should not be necessary anymore because we are always at least in C++14.

Please remove all changes to this file and switch from `_Void_t` to `void_t`

include/cuda/std/detail/libcxx/include/numeric
```diff
@@ -16,6 +16,7 @@
 namespace std
 {
 
+#ifndef __cuda_std__
```
miscco, line 19: Note to self, we should rather pull out the relevant pieces that we need (looking at https://github.com/llvm/llvm-project/tree/main/libcxx/include/__numeric)

That said, no chage requested for now

include/cuda/std/detail/libcxx/include/mdspan
```diff
@@ -0,0 +1,59 @@
...
+// Questions? Contact Christian R. Trott (crtrott@sandia.gov)
+//
+// ************************************************************************
+//@HEADER
+*/
+
+#ifndef _LIBCUDACXX_MDSPAN
+#define _LIBCUDACXX_MDSPAN
+
+#include "__mdspan/default_accessor.hpp"
```
miscco, line 47: I do not believe that this will work as intended. Please have a look at e.g the current iterators PR

https://github.com/NVIDIA/libcudacxx/pull/330/files#diff-fd3001206bd0e5e8adb45b08afdcaede30794e78ad7b57b7d39ee918c9e36c18R678-R733

include/cuda/std/detail/libcxx/include/__mdspan/type_list.hpp
```diff
@@ -0,0 +1,120 @@
...
+_LIBCUDACXX_BEGIN_NAMESPACE_STD
+
+//==============================================================================
+
+namespace detail {
+
+template <class... _Ts> struct __type_list { static constexpr auto __size = sizeof...(_Ts); };
+
+// Implementation of type_list at() that's heavily optimized for small typelists
+template <::std::size_t, class> struct __type_at;
```
miscco, line 60: Generally, you should not need to qualify `size_t` at all. This is a bit strange and suggest some missing include early on

include/cuda/std/detail/libcxx/include/__mdspan/trait_backports.hpp
```diff
@@ -0,0 +1,162 @@
...
+
+#ifdef __MDSPAN_NEEDS_TRAIT_VARIABLE_TEMPLATE_BACKPORTS
+
+#if __MDSPAN_USE_VARIABLE_TEMPLATES
+namespace std {
+
+#define __MDSPAN_BACKPORT_TRAIT(__TRAIT) \
+  template <class... _Args> __MDSPAN_INLINE_VARIABLE constexpr auto __TRAIT##_v = __TRAIT<_Args...>::value;
+
+__MDSPAN_BACKPORT_TRAIT(is_assignable)
```
miscco, line 66: Checking most of them I see them guarded by 
```cpp
#if _LIBCUDACXX_STD_VER > 11 && !defined(_LIBCUDACXX_HAS_NO_VARIABLE_TEMPLATES)
```

So I believe that is a remnant from a pre C++14 implementation?

include/cuda/std/detail/libcxx/include/__mdspan/trait_backports.hpp
```diff
@@ -0,0 +1,162 @@
...
+
+// </editor-fold> end Variable template trait backports (e.g., is_void_v) }}}1
+//==============================================================================
+
+//==============================================================================
+// <editor-fold desc="integer sequence (ugh...)"> {{{1
+
+#if !defined(__MDSPAN_USE_INTEGER_SEQUENCE) || !__MDSPAN_USE_INTEGER_SEQUENCE
+
+namespace std {
```
miscco, line 91: Should be `_LIBCUDACXX_BEGIN_NAMESPACE_STD`

include/cuda/std/detail/libcxx/include/__mdspan/trait_backports.hpp
```diff
@@ -0,0 +1,162 @@
...
+template <class _Tp, _Tp _Np, _Tp _Ip, _Tp... _Vals>
+struct __make_int_seq_impl<
+  _Tp, _Np, _Ip, integer_sequence<_Tp, _Vals...>
+> : __make_int_seq_impl<_Tp, _Np, _Ip+1, integer_sequence<_Tp, _Vals..., _Ip>>
+{ };
+
+} // end namespace __detail
+
+template <class _Tp, _Tp _Np>
+using make_integer_sequence = typename __detail::__make_int_seq_impl<_Tp, _Np, 0, integer_sequence<_Tp>>::type;
```
miscco, line 122: I believe those are alreayd available at C++14 so Could you check whether we actually need this file at all?

miscco (commented):
I am not fully through with the product code. 

That said, I want to reiterate, that I do not believe this is code is maintainable for us. I accept that we currently do not have capacity to write it from scratch and working code is definitely better than no code.

Nevertheless, I would want a commitment that we actually get the time to clean this up once we get the big features we are working on right now out of the window.

Before we can merge it we would need to remove all remnants of `std::` and make sure we are using the proper namespaces throughout.

include/cuda/std/detail/libcxx/include/__mdspan/macros.hpp
```diff
@@ -0,0 +1,652 @@
...
+    /**/
+#define __MDSPAN_TEMPLATE_REQUIRES_20(TP1, TP2, TP3, TP4, TP5, TP6, TP7, TP8, TP9, TP10, TP11, TP12, TP13, TP14, TP15, TP16, TP17, TP18, TP19, REQ) \
+  template<TP1, TP2, TP3, TP4, TP5, TP6, TP7, TP8, TP9, TP10, TP11, TP12, TP13, TP14, TP15, TP16, TP17, TP18, TP19 \
+    __MDSPAN_CLOSE_ANGLE_REQUIRES(REQ) \
+    /**/
+
+#define __MDSPAN_INSTANTIATE_ONLY_IF_USED \
+  __MDSPAN_TEMPLATE_REQUIRES( \
+    class __instantiate_only_if_used_tparam=void, \
+    ( __MDSPAN_TRAIT(::std::is_void, __instantiate_only_if_used_tparam) ) \
```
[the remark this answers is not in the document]
youyu3: Thanks - I thought that we wanted to use the one from `std` for some of them.
I checked how things like `is_void` are used in existing files (well should've checked earlier...).  Looks like, when it's already inside `_LIBCUDACXX_BEGIN_NAMESPACE_STD`/`_LIBCUDACXX_END_NAMESPACE_STD`, we just use `is_void` without a namespace.  And that'll automatically use the one in `_CUDA_VSTD`.  Would we prefer that?

include/cuda/std/detail/libcxx/include/__mdspan/trait_backports.hpp
mhoemmen: @miscco The macros wouldn't be original to the reference implementation, but there probably are some parts of the reference mdspan implementation that retain C++11 checks.  The reference mdspan implementation definitely requires C++14, though.

include/cuda/std/detail/libcxx/include/__mdspan/trait_backports.hpp
mhoemmen: @miscco This is probably a left-over from when the reference mdspan implementation had to support C++11.  It should be possible to remove this.

include/cuda/std/detail/libcxx/include/__mdspan/standard_layout_static_array.hpp
```diff
@@ -0,0 +1,688 @@
...
+#if __MDSPAN_PRESERVE_STANDARD_LAYOUT
+
+/**
+ *  PSA = "partially static array"
+ *
+ * @tparam _Tp
+ * @tparam _ValsSeq
+ * @tparam __sentinal
+ */
+template <class _Tag, class _Tp, class _static_t, class _ValsSeq, _static_t __sentinal = static_cast<_static_t>(dynamic_extent),
```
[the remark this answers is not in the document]
mhoemmen: @miscco I can never get a certain someone to spell "sentinel" right ; - ) .

include/cuda/std/detail/libcxx/include/__mdspan/macros.hpp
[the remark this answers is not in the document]
youyu3: Switched to `_CUDA_VSTD::`

include/cuda/std/detail/libcxx/include/__mdspan/macros.hpp
```diff
@@ -0,0 +1,652 @@
...
+
+#if __MDSPAN_USE_RETURN_TYPE_DEDUCTION
+#  define __MDSPAN_DEDUCE_RETURN_TYPE_SINGLE_LINE(SIGNATURE, BODY) \
+    auto __MDSPAN_PP_REMOVE_PARENS(SIGNATURE) { return __MDSPAN_PP_REMOVE_PARENS(BODY); }
+#  define __MDSPAN_DEDUCE_DECLTYPE_AUTO_RETURN_TYPE_SINGLE_LINE(SIGNATURE, BODY) \
+    decltype(auto) __MDSPAN_PP_REMOVE_PARENS(SIGNATURE) { return __MDSPAN_PP_REMOVE_PARENS(BODY); }
+#else
+#  define __MDSPAN_DEDUCE_RETURN_TYPE_SINGLE_LINE(SIGNATURE, BODY) \
+    auto __MDSPAN_PP_REMOVE_PARENS(SIGNATURE) \
+      -> std::remove_cv_t<std::remove_reference_t<decltype(BODY)>> \
```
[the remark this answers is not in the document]
youyu3: Done

include/cuda/std/detail/libcxx/include/__mdspan/macros.hpp
```diff
@@ -0,0 +1,652 @@
...
+
+template <class... _Args>
+constexpr __mdspan_enable_fold_comma __fold_comma_impl(_Args&&... __args) noexcept { return { }; }
+
+template <bool... _Bs>
+struct __bools;
+
+} // __fold_compatibility_impl
+
+} // end namespace std
```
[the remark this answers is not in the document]
youyu3: Done

include/cuda/std/detail/libcxx/include/__mdspan/macros.hpp
```diff
@@ -0,0 +1,652 @@
...
+#  define __MDSPAN_FOLD_AND_TEMPLATE(...) ((__VA_ARGS__) && ...)
+#  define __MDSPAN_FOLD_OR(...) ((__VA_ARGS__) || ...)
+#  define __MDSPAN_FOLD_ASSIGN_LEFT(__INIT, ...) (__INIT = ... = (__VA_ARGS__))
+#  define __MDSPAN_FOLD_ASSIGN_RIGHT(__PACK, ...) (__PACK = ... = (__VA_ARGS__))
+#  define __MDSPAN_FOLD_TIMES_RIGHT(__PACK, ...) (__PACK * ... * (__VA_ARGS__))
+#  define __MDSPAN_FOLD_PLUS_RIGHT(__PACK, ...) (__PACK + ... + (__VA_ARGS__))
+#  define __MDSPAN_FOLD_COMMA(...) ((__VA_ARGS__), ...)
+#else
+
+namespace std {
```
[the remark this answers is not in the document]
youyu3: Done

include/cuda/std/detail/libcxx/include/__mdspan/type_list.hpp
youyu3: All `::std::size_t` -> `size_t`

include/cuda/std/detail/libcxx/include/__mdspan/trait_backports.hpp
youyu3: This file if removed since we don't need C++11 support.

include/cuda/std/detail/libcxx/include/__mdspan/trait_backports.hpp
youyu3: This file if removed since we don't need C++11 support.

include/cuda/std/detail/libcxx/include/__mdspan/trait_backports.hpp
youyu3: This file is removed since we don't need C++11 support.

include/cuda/std/detail/libcxx/include/__mdspan/compressed_pair.hpp
```diff
@@ -0,0 +1,226 @@
...
+      : __t_val((_TLike &&) __t), __u_val((_ULike &&) __u) {}
+};
+
+#if !defined(__MDSPAN_USE_ATTRIBUTE_NO_UNIQUE_ADDRESS)
+
+// First empty.
+template <class _Tp, class _Up>
+struct __compressed_pair<
+    _Tp, _Up,
+    ::std::enable_if_t<__MDSPAN_TRAIT(::std::is_empty, _Tp) && !__MDSPAN_TRAIT(::std::is_empty, _Up)>>
```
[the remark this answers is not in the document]
youyu3: Done

include/cuda/std/detail/libcxx/include/__mdspan/config.hpp
youyu3: You mean replacing `#define __MDSPAN_HAS_CXX_14 (__MDSPAN_CPLUSPLUS >= __MDSPAN_CXX_STD_14)` (a few lines below here) with something like `#define __MDSPAN_HAS_CXX_14 (_LIBCUDACXX_STD_VER >= 14)`, and same thing for `__MDSPAN_HAS_CXX_17` and `__MDSPAN_HAS_CXX_20`?

include/cuda/std/detail/libcxx/include/__mdspan/compressed_pair.hpp
```diff
@@ -0,0 +1,226 @@
...
+  }
+  __MDSPAN_FORCE_INLINE_FUNCTION __MDSPAN_CONSTEXPR_14 _Up &__second() noexcept { return __u_val; }
+  __MDSPAN_FORCE_INLINE_FUNCTION constexpr _Up const &__second() const noexcept {
+    return __u_val;
+  }
+
+  __MDSPAN_INLINE_FUNCTION_DEFAULTED
+  constexpr __compressed_pair() noexcept = default;
+  __MDSPAN_INLINE_FUNCTION_DEFAULTED
+  constexpr __compressed_pair(__compressed_pair const &) noexcept = default;
```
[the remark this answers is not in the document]
mhoemmen: @miscco It's totally OK to review the implementation, but if we want to make major changes that diverge from the reference implementation, and if we think the reference implementation also needs correction, then would you consider please submitting an issue for the reference implementation?

include/cuda/std/mdspan
youyu3: Thanks.  Fixed.

include/cuda/std/detail/libcxx/include/type_traits
youyu3: I believe this came from the `std::span` PR.

miscco: You are absolutely right, I think it should already be gone from there

include/cuda/std/detail/libcxx/include/mdspan
youyu3: Thanks.  Added pragma_push and pragma_pop.

wmaxey (changes requested):
Commit history appears to be broken. I'll try to resolve this locally and possibly issue a correction, but AFAICTL there seems to be a replay of several commits.

youyu3:
> Commit history appears to be broken. I'll try to resolve this locally and possibly issue a correction, but AFAICT there seems to be a replay of several commits.
I pulled commits from the `span` PR.  There are further changes in those files I believe.

.upstream-tests/test/std/containers/views/mdspan/mdspan.accessor.default.members/access.pass.cpp
```diff
@@ -0,0 +1,24 @@
+//===----------------------------------------------------------------------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
```
wmaxey, line 8: All tests will likely need `//UNSUPPORTED: c++11` below the license and above the includes. 

This letds `lit` know that the tests are not designed for pre-C++14.

wmaxey: https://blossom.nvidia.com/sw-cuda-cccl-instance/blue/organizations/jenkins/cluster_launch_test/detail/cluster_launch_test/386/pipeline/1129

Results are crunchy, but it looks like there is maybe only a few failure points. 

I made a PR on gitlab that fixes the history. It drops the `<span>` PR and gets rid of some strange re-merge, but I think the content is identical.

youyu3: Thanks Wesley.  I can add the `//UNSUPPORTED: c++11` to the github PR.  Would you be able to merge that to the gitlab PR for CI?

youyu3: Also, what CUDA versions do our CI tests cover?

include/cuda/std/detail/libcxx/include/__mdspan/standard_layout_static_array.hpp
```diff
@@ -534,17 +534,9 @@ struct __standard_layout_psa<_Tag, _Tp, _static_t, _CUDA_VSTD::integer_sequence<
   //--------------------------------------------------------------------------
 
   __MDSPAN_INLINE_FUNCTION_DEFAULTED
-  constexpr __standard_layout_psa() noexcept
-#if defined(__clang__) || defined(__MDSPAN_DEFAULTED_CONSTRUCTORS_INHERITANCE_WORKAROUND)
```
mhoemmen: @youyu3 
Did the above change fix this issue, making this Clang work-around no longer needed?

libcxx/test/std/containers/views/mdspan/mdspan.extents.cons/span.pass.cpp
```diff
@@ -19,7 +21,7 @@ void test_span_con()
     using TestFixture = TestExtents<T>;
     TestFixture t;
 
-    auto s = std::span( t.dyn_sizes );
+    auto s = std::span<const size_t, t.dyn_sizes.size()>( t.dyn_sizes );
```
mhoemmen, line 24: @youyu3 If you're using `std::span`, you'll need C++20.  Would you consider using `cuda::std::span` (if it exists) instead?

youyu3: This test is for testing libcudacxx in the libcxx mode.  `std::span` is implemented in this mode.

include/cuda/std/detail/libcxx/include/__mdspan/standard_layout_static_array.hpp
youyu3: In my tests with clang, yes the work-around is no longer needed.

.upstream-tests/test/std/containers/views/mdspan/mdspan.extents.cons/array.pass.cpp
```diff
@@ -49,18 +49,18 @@ int main(int, char**)
     test_array_con< cuda::std::tuple_element_t< 4, extents_test_types > >();
     test_array_con< cuda::std::tuple_element_t< 5, extents_test_types > >();
 
-    static_assert( is_array_cons_avail_v< cuda::std::dextents<int,2>, int   , 2 > == true  );
+    static_assert( is_array_cons_avail_v< cuda::std::dextents<int,2>, int   , 2 > == true , "" );
```
mhoemmen, line 52: Just FYI, these are changes to support C++14 -- C++17 implements one-argument `static_assert`.

miscco (approved):

wmaxey (approved):

PR #300, opened 2022-08-10
Soundness bugfix for barrier<thread_scope_block> on sm_70

gonzalobg:
For sm_70, barrier arrive has an optimization to "coalesce" all arrives with the same update to the same barrier into a single update performed by a "leader" thread. 

This optimization is missing a release fence to establish cummulativity between all coalesced threads and the leader, before the leader performs the update.

### DIFF
--- a/include/cuda/std/barrier
+++ b/include/cuda/std/barrier
@@ -189,15 +189,17 @@ public:
                 : "r"(static_cast<std::uint32_t>(__cvta_generic_to_shared(&__barrier)))
                 : "memory");
 #else
-            unsigned int __activeA = __match_any_sync(__activemask(), __update);
-            unsigned int __activeB = __match_any_sync(__activemask(), reinterpret_cast<std::uintptr_t>(&__barrier));
+            unsigned int __mask = __activemask();
+            unsigned int __activeA = __match_any_sync(__mask, __update);
+            unsigned int __activeB = __match_any_sync(__mask, reinterpret_cast<std::uintptr_t>(&__barrier));
             unsigned int __active = __activeA & __activeB;
             int __inc = __popc(__active) * __update;
 
             unsigned __laneid;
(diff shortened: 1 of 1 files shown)


gonzalobg:
@daniellustig could maybe review?

include/cuda/std/barrier
```diff
@@ -189,15 +189,17 @@ public:
...
-            unsigned int __activeB = __match_any_sync(__activemask(), reinterpret_cast<std::uintptr_t>(&__barrier));
+            unsigned int __mask = __activemask();
+            unsigned int __activeA = __match_any_sync(__mask, __update);
+            unsigned int __activeB = __match_any_sync(__mask, reinterpret_cast<std::uintptr_t>(&__barrier));
             unsigned int __active = __activeA & __activeB;
             int __inc = __popc(__active) * __update;
 
             unsigned __laneid;
-            asm volatile ("mov.u32 %0, %laneid;" : "=r"(__laneid));
+            asm ("mov.u32 %0, %laneid;" : "=r"(__laneid));
```
griwes, line 199: Why are you removing the volatile here?

gonzalobg: Because it is not necessary. Two "calls" to this assembly instruction from this same thread always "return" the same value, so eliding one is a valid transformation for the compiler to do.

This assembly statement does not modify any memory, so it does not need a "memory" clobber either.

daniellustig:
The cumulativity fix seems reasonable to me

gonzalobg:
@wmaxey ?

davedsth (approved):
Looks ok, we should think of a refactor reinterpret_cast<std::uintptr_t>(&__barrier) and __match_any_sync

wmaxey:
Thanks for review, David. Merging.

PR #318, opened 2022-09-23
Fix gcc12 issues

miscco:
This fixes a host of gcc12 issues that are coming up in CI

Fixes nvbug3764574


.upstream-tests/test/std/utilities/time/time.clock/time.clock.system/local_time.types.pass.cpp
```diff
@@ -28,6 +28,7 @@
 
 int main(int, char**)
 {
+#ifndef __cuda_std__
```
griwes, line 31: This effectively disables this test - why?

miscco: because those types are guarded by `#ifndef __cuda_std__`

griwes: Ah, because we haven't done time zones. Cool.

griwes (approved):

wmaxey:
Squashing and merging! Thanks!

PR #319, opened 2022-09-28
Fix ubuntu18 failing in CI due to missing lit prereqs.

wmaxey:
Also reorder build steps in order of gravity.


miscco (approved):
Does nobody think about ubuntu18.04??

PR #321, opened 2022-10-05
[RFC] Make llvm a submodule

miscco:
Currently our base is copied by hand from libc++ source.

This is a rather cumbersome and error prone process that is prohibitively expensive for regular updates.

As a step forward, this RFC addes a submodule to our repo, that contains the llvm-project sources.

In order to bring in our local changes, it uses a dedicated branch. This is also one of the main curxes. Currently I just hardcoded my local fork of llvm. If we want to go this route, we would have to either create an official form of the llvm-project inside the nvidia organization or way better as the llvm-project, if we could add astable branch to theirs like `nvidia/libcudacxx_base`


docs/setup/getting.md
```diff
@@ -18,3 +18,19 @@ libcu++ is an open source project developed on GitHub, which is where you'll
...
 Our GitHub repository is [github.com/nvidia/libcudacxx](https://github.com/nvidia/libcudacxx).
 
+You can clone the repository via
+```bash
+git clone -c core.symlinks=true https://github.com/nvidia/libcudacxx
+```
+
+After cloning the repository, one needs to initialize the llvm submodule:
+```bash
+git submodule init
```
miscco, line 28: For whatever reasons, it is currently still not possible to sparsely checkout a submodule. So if a user wants to avoid pulling in all of llvm-project, then they will need to use this slightlöy cumbersome workaround

.clang-format
```diff
@@ -0,0 +1,84 @@
+BasedOnStyle: LLVM
```
miscco, line 1: Ignore that one please that slipped in

griwes:
As someone who was initially subtly opposed to removing submodules, I now believe that this is a step backwards. With a submodule, there's a significant amount of work you need to do every time you submit a change, reviewing PRs that touch files in both the main repo and in the submodule gets very annoying, exploring history of the changes is not trivial, and there's internal perforce/CUDA installer considerations that are somewhat complicated.

I would like to see what advantages of moving backwards to a submodule solution you're seeing that outweight the above problems. Keep in mind that we desubmodularized after discussions with upstream where they said they will actively purge code for old compiler support (and by that they really mean, "not clang trunk") from the upstream repo, which means we cannot automatically pull from upstream without huge amounts of work to undo their changes that break our support matrix every time.

miscco:
> As someone who was initially subtly opposed to removing submodules ....

This is actually meant as a RFC exploring possible ways forward and your comments are very valuable.

Currently we are far removed from the original source of libc++.

This makes it increasingly more expensive to update the sources as time goes by. I believe that submodules would not help that, but would allow us to better keep up to date once we have reached a less far removed state.

Is that worth it? I am not sure and would currently also slightly favor staying where we are.

miscco:
After internal discussions this is not something we want to pursue in the near future

PR #329, opened 2022-11-03
Block local memory tests on pascal

wmaxey:
Pascal cannot use the general scoped atomics for `.local` memory, I don't know if it can use the `.local` scoped atomics to avoid this issue which is our next planned refactor. This is not the case for SM_70 and doesn't appear to be documented anywhere. This PR then changes all the tests to instead exclusively test shared/global memory on Pascal.

PR makes Pascal tests happy which in turn makes CI happy.
https://blossom.nvidia.com/sw-cuda-cccl-instance/blue/organizations/jenkins/cluster_launch_test/detail/cluster_launch_test/311/pipeline/1062


.upstream-tests/test/std/atomics/atomics.types.generic/atomic_copyable.pass.cpp
```diff
@@ -44,21 +45,23 @@ void test_copy_constructible() {
...
   T t0(val);
   T t1(t0);
 
-  t0++;
-  t1++;
+  // t0++;
+  // t1++;
 
-  assert(t1.load() == 2);
+  // assert(t1.load() == 2);
```
griwes, line 57: I assume that these weren't meant to be left commented out?

griwes (approved):
LGTM other than the one comment.

.upstream-tests/test/std/atomics/atomics.types.generic/atomic_copyable.pass.cpp
wmaxey: Fixed! Thanks.

.upstream-tests/test/std/atomics/atomics.lockfree/isalwayslockfree.pass.cpp
```diff
@@ -18,9 +18,13 @@
 
 #include "test_macros.h"
 
+// NVRTC doesn't include host atomic making this feature test invalid
+// TODO: Should we define __cpp_lib_atomic_is_always_lock_free for NVRTC?
```
wmaxey, line 22: @griwes Food for thought regarding macro. We don't do this now for anything so it's going to have to be another PR.

miscco (approved):

PR #330, opened 2022-11-04
Modularize <iterator>

miscco:
This modularizes the `<iterator>` header as done by libc++.

This is a precursor to  updating to a more recent state of the code and bringing in C++20 ranges machinery


.upstream-tests/test/std/iterators/iterator.container/ssize.pass.cpp
```diff
@@ -142,7 +143,9 @@ int main(int, char**)
     static_assert( cuda::std::is_signed_v<                      decltype(cuda::std::ssize(sc))>, "");
     static_assert( cuda::std::numeric_limits<                   decltype(cuda::std::ssize(sc))>::max()  > 60000, "");
     static_assert( cuda::std::numeric_limits<cuda::std::make_signed_t<decltype(cuda::std:: size(sc))>>::max() < 60000, "");
+#ifdef __CUDA_ARCH__
```
griwes, line 147: Why?

include/cuda/std/detail/libcxx/include/__iterator/reverse_iterator.h
```diff
@@ -0,0 +1,186 @@
...
+template <class _Iter>
+class _LIBCUDACXX_TEMPLATE_VIS reverse_iterator
+    : public iterator<typename iterator_traits<_Iter>::iterator_category,
+                      typename iterator_traits<_Iter>::value_type,
+                      typename iterator_traits<_Iter>::difference_type,
+                      typename iterator_traits<_Iter>::pointer,
+                      typename iterator_traits<_Iter>::reference>
+{
+private:
+    /*mutable*/ _Iter __t;  // no longer used as of LWG #2360, not removed due to ABI break
```
griwes, line 49: Not really all that relevant to this PR, but since you've made me look through all this code: I wonder why this isn't changed with unstable ABI and if we should bite the bullet and remove it under unstable ABI in our copy.

.upstream-tests/test/std/iterators/predef.iterators/move.iterators/move.iter.ops/move.iter.op.index/difference_type.pass.cpp
```diff
@@ -46,6 +46,9 @@ int main(int, char**)
 {
     {
         char s[] = "1234567890";
+#ifdef __NVCOMPILER
```
griwes, line 50: Why? (I assume a compiler bug, but we should leave an explanation for this one here.)

griwes (approved):
Minor comments only, ship it.

.upstream-tests/test/std/iterators/predef.iterators/move.iterators/move.iter.ops/move.iter.op.index/difference_type.pass.cpp
miscco: Yeah there is a bug. That said it is so fundamental, that it shouldnt be a compiler bug.

In a nutshell with nvc++ the array s gets initialized to char s[] = "1111111111";

Why I have absolutely no idea. Given that it works on godbolt I assume something fishy with our test infra

.upstream-tests/test/std/iterators/iterator.container/ssize.pass.cpp
miscco: This is again some very strange compiler bug that leads to wraparound on device

I should have added comments to those 🤦

include/cuda/std/detail/libcxx/include/__iterator/reverse_iterator.h
miscco: We should consider that in the future, but not in this PR

.upstream-tests/test/std/iterators/predef.iterators/move.iterators/move.iter.ops/move.iter.op.index/difference_type.pass.cpp
griwes: We should get to the bottom of this before merging this PR then.

miscco: That is preexisting, so I do not think it should prevent merging this

miscco: To be clear, in the current test config the following fails on device only
```cpp
int main(int, char**)
{
    char s[] = "1234567890";
    assert(s[5] == 6);
    return 0;
}
```

I do not know whether i configured it wrong, but nvc++ does some strange things

griwes: Okay, I didn't realize this was preexisting. We should figure this out fully, but I'm ok not blocking the PR over it.

PR #331, opened 2022-11-08
Correct issues with CMake install rules

robertmaynard:
This corrects three issues found by `rapids-cmake` and others when trying to install libcudacxx.

1. The `install` commands need to come after `project`. The `install` commands aren't allowed when in CMake scripting mode which is the mode you are in before the first call to `project()`
2. The `install(DIRECTORY PATTERN)` matches the end of a file name, not the front so we need to use `*.cmake.in` to not
install the template files
3. We incorrectly used `libcudacxx_TOPLEVEL_PROJECT` instead of `LIBCUDACXX_TOPLEVEL_PROJECT` so we never defaulted to have install rules enabled when building libcudacxx itself.
4. Don't install any CMakeLists.txt files that are inside `include/cuda' or `include/nv'

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,6 +11,11 @@ else()
   set(LIBCUDACXX_TOPLEVEL_PROJECT OFF)
 endif()
 
+set(PACKAGE_NAME libcudacxx)
+set(PACKAGE_VERSION 11.0)
+set(PACKAGE_STRING "${PACKAGE_NAME} ${PACKAGE_VERSION}")
+project(libcudacxx NONE)
+
 include(cmake/libcudacxxInstallRules.cmake)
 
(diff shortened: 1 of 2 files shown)


miscco (approved):
LGTM but I feel like a 3-year old whenever I see cmake, so I would love some feedback from more knowledgeable folks

wmaxey:
Pipeline has been launched on Blossom. We'll know the results soon.

robertmaynard:
> Pipeline has been launched on Blossom. We'll know the results soon.

Any update on this?

allisonvacanti (approved):
lgtm

wmaxey:
> > Pipeline has been launched on Blossom. We'll know the results soon.
> 
> Any update on this?

Pipeline was having issues over the weekend and earlier. Tests are passing. LGTM.

PR #336, opened 2022-12-01
Fetch host `<utility>` to populate `tuple_size`

wmaxey:


miscco:
Could you also then remove the forward declarations inside of array and tuple?

miscco (approved):

jrhemstad (commented):
Don't we need to add `#include <utility>` to `cuda/std/array` too?

miscco:
`<array>` already includes `<utility>`

PR #338, opened 2022-12-06
<complex> Implement p0415r1 `constexpr For <complex> (Again)`

miscco:
This implements p0415r1 `constexpr For <complex> (Again)` which can be found here https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2017/p0415r1.html

Unfortunately we cannot currently implement it fully because there are some issues with intrinsics:

1. __isinf is currently not constexpr on device side. We need compiler support for that.
2. Divide is calling a lot of non constexpr functions. We would have to rewrite the algorithm completely :(

I did add a copyright banner to all relevant test cases when comparing them to the current state of libc++


miscco:
@jrhemstad Do we want to get this in even given that we only partially support the feature?

PR #343, opened 2022-12-30
Fix missing definition of ASSERT functionality

miscco:
We missed somehow to properly include the assert functionality everywhere

To remedy that port the new modularized assert definitions from libc++

I am slightly scared by the config changes but I guess we cannot really work around this forever


miscco:
Dropping this because it is integrated into a different PR

PR #345, opened 2022-12-30
Fix warning about unqualified move

miscco:
This came up with newer clang releases

fixes nvbug3918987


mhoemmen (approved):
I didn't realize we had a `valarray` implementation!  Anyway, thanks for the fix!

wmaxey:
> I didn't realize we had a `valarray` implementation! Anyway, thanks for the fix!

This is for libcxx tests. It's not exposed in libcu++.

wmaxey (approved):

PR #347, opened 2023-01-04
Fix errors in atomic with small aggregates and enum classes

wmaxey:


include/cuda/std/detail/libcxx/include/support/atomic/atomic_cuda.h
```diff
@@ -121,13 +121,43 @@ const volatile _Tp* __cxx_get_underlying_device_atomic(__cxx_atomic_base_heterog
 }
 
 template <typename _Tp>
-using __cxx_atomic_small_to_32 = typename conditional<is_signed<_Tp>::value, int32_t, uint32_t>::type;
+using __cxx_atomic_small_to_32 =
+    typename conditional<is_signed<_Tp>::value,
+        int32_t,
+        uint32_t
+    >::type;
```
miscco, line 128: We have a helper alias for this now, so that we do not have to bother with the `meow_t` not being available in C++11
```suggestion
using __cxx_atomic_small_to_32 = __conditional_t<is_signed<_Tp>::value, int32_t, uint32_t>;
```

include/cuda/std/detail/libcxx/include/support/atomic/atomic_cuda.h
```diff
@@ -121,13 +121,43 @@ const volatile _Tp* __cxx_get_underlying_device_atomic(__cxx_atomic_base_heterog
...
+
+// Arithmetic conversions to/from proxy types
+template<class _Tp, _EnableIf<is_arithmetic<_Tp>::value, int> = 0>
```
miscco, line 131: Ditto, we should move to
```suggestion
template<class _Tp, __enable_if_t<is_arithmetic<_Tp>::value, int> = 0>
```

include/cuda/std/detail/libcxx/include/support/atomic/atomic_cuda.h
```diff
@@ -121,13 +121,43 @@ const volatile _Tp* __cxx_get_underlying_device_atomic(__cxx_atomic_base_heterog
...
+constexpr __host__ __device__ inline __cxx_atomic_small_to_32<_Tp> __cxx_small_to_32(_Tp __val) {
```
miscco, line 132: Should this rather use `_LIBCUDACXX_INLINE_VISIBILITY`. Also I am not sure if we need the inline here,a s the function is already constexpr
```suggestion
_LIBCUDACXX_INLINE_VISIBILITY constexpr __cxx_atomic_small_to_32<_Tp> __cxx_small_to_32(_Tp __val) {
```

miscco (commented):
Some minor nits about internal helpers. (Allies throughout product code)

include/cuda/std/detail/libcxx/include/support/atomic/atomic_cuda.h
wmaxey: I was under the impression that the meta base stuff was better to use internally.

include/cuda/std/detail/libcxx/include/support/atomic/atomic_cuda.h
wmaxey: Updated.

include/cuda/std/detail/libcxx/include/support/atomic/atomic_cuda.h
miscco: That is correct, but my evil plan is to settle on `__enable_if_t` globally and then move that over to use the meta based implementation. I really do not like it that we have three different ways to do this.

So anything here is fine by me as that should be a separate PR anyway

miscco: That is really strange, do we build in C++03?

miscco:
:facepalm: https://github.com/NVIDIA/libcudacxx/blob/25658c1bc5b93f58715ff893cda2643e85c317d4/include/cuda/std/detail/libcxx/include/__type_traits/conditional.h#L46-L54

Seems I put it inside the ifdef :(

wmaxey:
> https://github.com/NVIDIA/libcudacxx/blob/25658c1bc5b93f58715ff893cda2643e85c317d4/include/cuda/std/detail/libcxx/include/__type_traits/conditional.h#L46-L54
> 
> Why is it not there?

Perhaps I just need to update the branch it's based on. I'll do that today.

include/cuda/std/detail/libcxx/include/support/atomic/atomic_cuda.h
wmaxey: I rebased the branch, which I know is fairly aggressive, but now all the conditionals are using the `__blah_t` versions.

miscco (approved):

PR #352, opened 2023-01-18
Make lerp usable on device

miscco:
This currently breaks our CI with nvrtc


wmaxey (approved):
This was approved earlier in another PR.

PR #356, opened 2023-01-23
Implement ranges machinery from <iterator>

miscco:
This implements most of the ranges machinery that is needed for e.g `contiguous_iterator` and also modernizes the rest of the iterator headers to the current state of C++20


miscco:

If I get a unicorn and a pony we can always reopen it

PR #357, opened 2023-01-23
Implement range concepts

miscco:
This implements all the machinery related to what a range actually is and prepares the way for views.

This includes the CPOs as well as the concepts


miscco:

PR #361, opened 2023-02-04
Add support for testing on ARM64

wmaxey:
Hacky, but by setting the environment up to override the default setting we can now test on ARM.

```
DEFAULT_ARCH=linux/arm64
docker compose --profile gcc-12 -f "environments\linux\docker\compose.yml" build
docker run -it --gpus all libcudacxx/gcc-12
```
Result:
```
PS C:\Users\Wes\Documents\workspace\libcudacxx> docker run -it --gpus all libcudacxx/gcc-12
WARNING: The requested image's platform (linux/arm64) does not match the detected host platform (linux/amd64) and no specific platform was requested

root@477ca5b56973:/#
```

Obviously if you are running on x86, you can't passthrough a GPU to the guest image. That's fine for `NoopExecutor()` testing, but this lets us begin testing on ARM machines internally.

### DIFF
--- a/environments/linux/docker/.env
+++ b/environments/linux/docker/.env
@@ -1,0 +1,2 @@
+### Set default arch
+DEFAULT_ARCH=linux/amd64
--- a/environments/linux/docker/compose.yml
+++ b/environments/linux/docker/compose.yml
@@ -4,6 +4,7 @@ services:
     image: "local/cccl-ubuntu22.04"
     tty: true
     profiles: ["base"]
+    platform: $DEFAULT_ARCH
     build:
       cache_from:
         - "local/cccl-ubuntu22.04:latest"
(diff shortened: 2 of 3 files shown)


environments/linux/docker/.env
```diff
@@ -0,0 +1,2 @@
+### Set default arch
+DEFAULT_ARCH=linux/amd64
```
miscco, line 2: Could we determine this via `dpkg --print-architecture`

wmaxey: Unless we make a script to launch the docker compose bits, I'd rather see the runner actually request the different arch with `DEFAULT_ARCH=linux/arm64 docker compose blahhh`. 

This allows QA to make changes where they need to.

miscco (approved):

griwes (approved):

wmaxey:
I'm closing this as the work on devcontainers supersedes the need for these changes.

PR #367, opened 2023-02-17
Fix warnings in several headers

wmaxey:
MSVC is *extremely* pedantic with `/Wall` enabled so we'll skip that entirely.


wmaxey:
This fixes #297. I made some effort to find out why that warning wasn't being issued in the first place, and it's because we use `-Wno-unused-parameters` to sink a ton of warnings across the test files themselves. This would be a pretty large fixup that can be done at a later time.

The actual fix for #297 on the other hand is to add the system header pragma to some of the files that need it. Specifically files that implement additional things on top of `cuda/std/`.

.upstream-tests/test/CMakeLists.txt
```diff
@@ -37,11 +37,19 @@ else() # NOT LIBCUDACXX_TEST_WITH_NVRTC
...
-set(LIBCUDACXX_TEST_COMPILER_FLAGS
-  "${LIBCUDACXX_TEST_COMPILER_FLAGS} \
-  ${LIBCUDACXX_FORCE_INCLUDE} \
-  -I${CMAKE_SOURCE_DIR}/include"
-  CACHE INTERNAL "Flags for libcxx testing." FORCE)
+if (NOT MSVC)
+  set(LIBCUDACXX_WARNING_LEVEL
+    "--compiler-options=-Wall \
+     --compiler-options=-Wextra")
+endif()
```
miscco, line 44: I am not too happy with this. That said, we can reenable them once we have a workaround in place

miscco (approved):

PR #369, opened 2023-02-20
Fix cuda::atomic_thread_fence example

senior-zero:
Atomic operations are still requires to establish synchronization order in the pretense of the `std::atomic_thread_fence`. This PR provides a less ambiguous example of the API.


gonzalobg (approved):
LGTM, thanks!

miscco (approved):

PR #371, opened 2023-02-21
Document asynchronous threads

gonzalobg:


gonzalobg:
@daniellustig for review.

miscco (approved):

gonzalobg:
Let’s wait for Dan’s feedback since I’m not sure at all about this one. Just opened a new PR because the old one got quite out of sync with respect to the latest changes in the repo.
________________________________

docs/extended_api/memory_model.md
```diff
@@ -39,6 +46,7 @@ Each program thread is related to each other program thread by one or more threa
 - Each GPU thread is related to each other GPU thread in the same CUDA device by the *device* thread scope: `thread_scope_device`.
 - Each GPU thread is related to each other GPU thread in the same CUDA thread block by the *block* thread scope: `thread_scope_block`.
 - Each thread is related to itself by the `thread` thread scope: `thread_scope_thread`.
+- Asynchronous threads are related to the thread requesting the asynchronous operations via all scope relationships.
```
daniellustig, line 49: Minor nit with my language lawyer hat on: the standard seems to use "create thread" rather than "request".  Maybe this instead? `Each thread is related to each asynchronous thread that it creates by all scopes.`

gonzalobg: I could not find a definition of "create thread" operation in the standard. It seems like the parallel algorithms _assume_ that "create thread" has the same _Synchronization_ properties as the constructor of `std::thread` and `std::jthread` (http://eel.is/c++draft/thread.thread.constr#6):

> The completion of the invocation of the constructor synchronizes with the beginning of the invocation of the copy of f[.](http://eel.is/c++draft/thread.thread.constr#6.sentence-1)

Do we have a note somewhere documenting that "Creating an asynchronous thread does not synchronize the paraent thread with the child thread."? If not, should I add that Note right here?

jrhemstad: @daniellustig thoughts on Gonzalo's question above?

daniellustig: I don't follow, why would we say that it does not synchronize?

gonzalobg: The C++ thread creation APIs establish memory ordering between the parent and child threads. 
If we say that creating the async thread synchronizes, I'd understand that as the parent thread executes a release-pattern before creating the thread, and the async thread performs an acquire-patter that synchronizes with the release-pattern of the parent, before doing anything else. 

I think that this is what we want and need, so saying create sounds fine: we want cumulativity on the async thread creation side, and only want to loose it on the arrival side.

gonzalobg:
Ready to merge @daniellustig

hcedwar:
Initiation of an async-operation (memcpy_async) is not _required_ to spawn a new async-thread.  It synchronizes-with an async-thread (_could_ spawn) which is responsible for eventual completion of the async-operation.

miscco (approved):
I believe that the `__as-if__` in the wording is sufficient to make clear that we do not in fact spawn a new thread

gonzalobg:
Making it a draft for the time being.

miscco:
Could you also please reopen this against the cccl repo?

PR #384, opened 2023-03-09
Move the `unused` helper function into `test_macros.h`

miscco:


wmaxey:
This is only used 35 times? I added a use of it, but will edit it in this PR after a rebase.

wmaxey (approved):

PR #385, opened 2023-03-09
Update documentation of cuda::discard_memory

gonzalobg:


gonzalobg:
@daniellustig for review.

docs/extended_api/memory_access_properties/discard_memory.md
```diff
@@ -12,11 +12,9 @@ __device__ void discard_memory(void volatile* ptr, size_t nbytes);
 
 **Preconditions**: `ptr` points to a valid allocation of size greater or equal to `nbytes`.
 
-**Effects**: equivalent to `memset(ptr, _indeterminate_, nbytes)`.
+**Effects**: `discard_memory` is a weak memory operation that behaves like a write for the purpose of conflicting accesses. After executing it, a value must be written to the memory before reading from it; otherwise, the behavior is _undefined_.
```
daniellustig, line 15: I actually think the original definition is more precise.  An evaluation that produces an indeterminate value is UB: https://eel.is/c++draft/basic#indet-2, so any read that races with the discard is UB by that rule.  And "before" isn't precise in the text as written; it would have to be "happens before" or something.

So maybe keep that original effects definition, and move the proposed new second sentence to follow the existing note below?  That would be a good place for the more intuitive informal explanation.

PR #388, opened 2023-03-10
Fix `_EnableConstructor` to be dependent on a template argument

miscco:
so that clang is happy


senior-zero (approved):

miscco:
Also a drive by fix to mark memory_resource as unsupported under windows for now

wmaxey (approved):

PR #389, opened 2023-03-10
Fix atomic_ref example in docs

sleeepyjack:
The example shown in the docs of `atomic_ref` is incorrect. The associated Godbolt link, however, is correct.


miscco (approved):

miscco:
Thanks a lot 🎉

PR #397, opened 2023-03-17
Implement P04151 constexpr for <complex> (Again)

miscco:
This implements the constexpr extensions for complex that made it into C++20. We also backport them to C++14 because there is no real reason to not do that.

In addition to the changes in the paper, we also add interoperability between `cuda::std::complex` and `std::complex`

This allows us to later replace `thrust::complex` with `cuda::std::complex`

Finally we enable users to skip some expensive normalization steps of complex multiplication that would result in a denormal


wmaxey:
MSVC is entirely green on 14.33+ and C++20. There appears to be some amount of code emitted by NVCC that breaks MSVC's complex header in C++20...

The repro is just `#include <complex>` so not likely our problem, but our inclusion of it due to the compatibility with `std::complex` causes all of our tests to fail. I'll see if this is a known issue.

```
C:\msbuild\2022\VC\Tools\MSVC\14.32.31326\include\complex(549): error C3615: constexpr function 'std::_Ctraits<long double>::_Isnan' cannot result in a constant expression
C:\msbuild\2022\VC\Tools\MSVC\14.32.31326\include\complex(553): note: failure was caused by call of undefined function or one not declared 'constexpr'
C:\msbuild\2022\VC\Tools\MSVC\14.32.31326\include\complex(553): note: see usage of 'std::_Bit_cast'
C:\msbuild\2022\VC\Tools\MSVC\14.32.31326\include\complex(688): error C3615: constexpr function 'std::_Ctraits<double>::_Isnan' cannot result in a constant expression
C:\msbuild\2022\VC\Tools\MSVC\14.32.31326\include\complex(689): note: failure was caused by call of undefined function or one not declared 'constexpr'
C:\msbuild\2022\VC\Tools\MSVC\14.32.31326\include\complex(689): note: see usage of 'std::_Bit_cast'
C:\msbuild\2022\VC\Tools\MSVC\14.32.31326\include\complex(826): error C3615: constexpr function 'std::_Ctraits<float>::_Isnan' cannot result in a constant expression
C:\msbuild\2022\VC\Tools\MSVC\14.32.31326\include\complex(827): note: failure was caused by call of undefined function or one not declared 'constexpr'
C:\msbuild\2022\VC\Tools\MSVC\14.32.31326\include\complex(827): note: see usage of 'std::_Bit_cast'
```

wmaxey (approved):

.upstream-tests/test/std/numerics/complex.number/complex/abi_v3.pass.cpp
```diff
@@ -1,44 +0,0 @@
-//===----------------------------------------------------------------------===//
```
wmaxey, line 1: Why delete this file?

miscco: Urgh, good catch, I replaced our folders with upstream to see the changes in the tests and seem to have forgotten to restore those

PR #400, opened 2023-03-20
Avoid duplicate symbol definition

miscco:
In some of the backported features we were accidentally using inline variables unconditionally.

This trips up the modules tests, as those variable templates would technically define duplicated symbols per inclusion of the header in C++14.

To work around this and keep efficient code c++17 onwards move the `__MDSPAN_TRAIT` macro into `__config` and use it when applicable


wmaxey (approved):

include/cuda/std/detail/libcxx/include/__concepts/swappable.h
```diff
@@ -112,6 +85,29 @@ _LIBCUDACXX_BEGIN_NAMESPACE_CPO(__swap)
   _LIBCUDACXX_CONCEPT __exchangeable = _LIBCUDACXX_FRAGMENT(__exchangeable_, _Tp);
 #endif // _LIBCUDACXX_STD_VER < 20
 
+
+#if _LIBCUDACXX_STD_VER > 17 && !defined(_LIBCUDACXX_COMPILER_PGI) // nvbug4051640
+  struct __fn;
+
+#if defined(_LIBCUDACXX_COMPILER_NVCC)
+#  pragma nv_diag_suppress 2642
```
miscco, line 93: @wmaxey I have removed the CUDART version check as C++20 requires C++20 so we need at least 12.0

senior-zero (approved):

PR #406, opened 2023-03-24
Relax CMake package compatibility checks

senior-zero:
Thrust/CUB CMake is currently searching for libcu++ with a minimal version of 1.8 or higher. On the libcu++ side, CMake is configured such that major versions should match exactly. This leads to Thrust/CUB rejecting libcu++ 2.1:
```
[cmake] CMake Warning at cub/cmake/cub-config.cmake:64 (find_package):
[cmake]   Could not find a configuration file for package "libcudacxx" that is
[cmake]   compatible with requested version "1.8.0".
``` 
This PR relaxes mentioned constrain, allowing users to find libcu++ with a greater major version.


miscco (approved):

robertmaynard:
Is it presumed that going forward libcudacxx will offer compat across major version values? If not we should encode this as an exception that 1.X and 2.X are compat.

wmaxey:
> Is it presumed that going forward libcudacxx will offer compat across major version values? If not we should encode this as an exception that 1.X and 2.X are compat.

As I understand it, the answer is yes. We might deprecate cuda extensions or change ABI, but the standards defined functionality should always be compatible. It's certainly up for discussion.

What would that look like? Do we check version greater than 1.X?

robertmaynard:
> What would that look like? Do we check version greater than 1.X?

Yeap!

lib/cmake/libcudacxx/libcudacxx-config-version.cmake
```diff
@@ -26,7 +26,7 @@ set(PACKAGE_VERSION_EXACT FALSE)
 set(PACKAGE_VERSION_UNSUITABLE FALSE)
 
 if(PACKAGE_VERSION VERSION_GREATER_EQUAL PACKAGE_FIND_VERSION)
-  if(PACKAGE_FIND_VERSION_MAJOR VERSION_EQUAL libcudacxx_VERSION_MAJOR)
+  if(PACKAGE_FIND_VERSION_MAJOR VERSION_LESS_EQUAL libcudacxx_VERSION_MAJOR)
```
wmaxey, line 29:
```suggestion
  if(PACKAGE_FIND_VERSION_MAJOR VERSION_GREATER_EQUAL 1)
```

@robertmaynard I feel like this was a silly thing to write out, but if we ever decide `2.1.0` is our minimum between CUB and Thrust we can change this.

PR #410, opened 2023-03-27
Add support for clang-16

wmaxey:


miscco (approved):
I removed the Intel specific workarounds at the end of the file as those builtins are now universally disabled

PR #411, opened 2023-03-28
Support an NVHPC config

wmaxey:
There appears to be a deadlock in semaphore when compiling with  NVC++. It's tricky and happens occasionally.

In order to not completely kill testing, I've cobbled together the timeout stuff into our config.


miscco (approved):
Please double check, there were some incoming `_LIBCUDACXX_COMPILER_PGI` recently

wmaxey:
> Please double check, there were some incoming `_LIBCUDACXX_COMPILER_PGI` recently

Do you mean there are more? I replaced everything I found.

miscco:
> > Please double check, there were some incoming `_LIBCUDACXX_COMPILER_PGI` recently
> 
> Do you mean there are more? I replaced everything I found.

I merged main and replaced the newly added occurences

.upstream-tests/test/support/test_macros.h
```diff
@@ -62,7 +62,7 @@
 #define TEST_HAS_BUILTIN_IDENTIFIER(X) 0
 #endif
 
-#if defined(__PGIC__)
+#if defined(__NVHPC__)
 # define TEST_COMPILER_PGI
```
miscco, line 66: I think you need to replace that one too ^^

include/cuda/std/detail/libcxx/include/cmath
```diff
@@ -308,6 +308,10 @@ long double    truncl(long double x);
 #include <math.h>
 #endif
 
+#if defined(__cuda_std__) && defined(_LIBCUDACXX_COMPILER_NVHPC)
```
wmaxey, line 311: https://clang.godbolt.org/z/hMWf98ae1

There is some weirdness depending on the toolchain NVC++ uses. Not sure how we need to configure the NVHPC image to get it to emit some of these functions.

miscco: I am strongly against us having to work around missing functionality. We should rather ask where those can be found.

wmaxey: Fixed after consultation with David.

PR #414, opened 2023-03-30
Unbreak `__make_unsigned` for clang-16

miscco:
clang-16 introduced a builtin `__make_unsigned` which unfortunately clashes with a name defined by libstdc++ so we cannot ever use it.

To add insult to injury our fallback struct was also named `__make_unsigned` which clang complains about because it clashes with the builtin.

To alleviate this rename `__make_unsigned` to `__make_unsigned_impl`


senior-zero (approved):
Since the conflict is resolved, are we allowed to return the following now?
```cpp
// Disabled due to libstdc++ conflict
#if 0 // __check_builtin(make_signed)
#define _LIBCUDACXX_MAKE_SIGNED(...) __make_signed(__VA_ARGS__)
#endif

// Disabled due to libstdc++ conflict
#if 0 // __check_builtin(make_unsigned)
#define _LIBCUDACXX_MAKE_UNSIGNED(...) __make_unsigned(__VA_ARGS__)
#endif
```

miscco:
> Since the conflict is resolved, are we allowed to return the following now?
> 
> ```c++
> // Disabled due to libstdc++ conflict
> #if 0 // __check_builtin(make_signed)
> #define _LIBCUDACXX_MAKE_SIGNED(...) __make_signed(__VA_ARGS__)
> #endif
> 
> // Disabled due to libstdc++ conflict
> #if 0 // __check_builtin(make_unsigned)
> #define _LIBCUDACXX_MAKE_UNSIGNED(...) __make_unsigned(__VA_ARGS__)
> #endif
> ```

Oh this is the other way around. gcc defines a function with that name, which clashes with the builtin. That is why we need to disable using the builtin.

However, at the same time the fallback defines a struct which has the same name of the builtin, which clang-16 still recognizes so it complains that we misuse a reserved keyword

PR #418, opened 2023-03-30
Use the CUDA libc++ thread API when compiling with Emscripten.

brycelelbach:
As pthreads isn't available in Emscripten/WASM, libcu++ fails to build there, complaining that no thread API is found. We need to use the CUDA libc++ thread API for this environment, just as we do for NVRTC. This fixes https://github.com/NVIDIA/thrust/issues/1902.


wmaxey:
I'm happy to merge this as is, and there will be a 2.1.1 release soon for libcudacxx that this can be a part of.

Long term we should figure out how to add tests for this. Even if they're just smoky config checks.

wmaxey (approved):

PR #425, opened 2023-04-06
Rework our __config

miscco:
Our config file is what one would call organically grown. To add insult to injury we have two of them

This consolidates a lot of the config into the one inside libcxx and uses `__cuda_std__` to differentiate between the different modes.

With the exception of the one commit that moves the builtins, everything is put into really small commits that should be easy to review


miscco:
Our Ci run looks reasonably well https://blossom.nvidia.com/sw-cuda-cccl-instance/blue/organizations/jenkins/cluster_launch_test/detail/cluster_launch_test/693/pipeline/111/

include/cuda/std/detail/libcxx/include/__cuda/barrier.h
```diff
@@ -23,6 +23,12 @@
 #pragma GCC system_header
 #endif
 
+#if defined(_LIBCUDACXX_COMPILER_NVRTC)
+#define _LIBCUDACXX_OFFSET_IS_ZERO(type, member) !(&(((type *)0)->member))
```
wmaxey, line 27: I wonder if this was fixed in nvrtc. 🤔

include/cuda/std/detail/libcxx/include/__cuda/climits_prelude.h
```diff
@@ -15,11 +15,11 @@
 #error "<__cuda/climits_prelude> should only be included in from <cuda/std/climits>"
 #endif // __cuda_std__
 
-#ifndef __CUDACC_RTC__
+#ifndef _LIBCUDACXX_COMPILER_NVRTC
     #include <climits>
     #include <limits.h>
     #include <cstdint>
-#else
+#else // ^^^ !_LIBCUDACXX_COMPILER_NVRTC ^^^ / vvv _LIBCUDACXX_COMPILER_NVRTC vvv
```
wmaxey, line 22: I wonder if we can lint this across the code base. It would make the config much more fun to read.

include/cuda/std/detail/__config
```diff
@@ -45,19 +45,6 @@
 #endif
 
 // request these outcomes
-
-#if defined(__CUDACC_RTC__)
-#  if defined(__CUDACC_RTC_INT128__)
-#    define __SIZEOF_INT128__ 16 // This macro is required in order to use int128 within the library
```
wmaxey, line 62: Why did this macro exist. 🤔 I feel like this was some hack job leftover that I wrote back when NVCC initially added support.

wmaxey:
I reviewed this commit by commit, this looks good! I appreciate that thread API handling is more permissive now, but I would like to take a look over `<atomic>` and see if we can warn when we encounter compilers outside of our test matrix. Looking at QCC and IBM AIX specifically.

wmaxey (approved):

include/cuda/std/detail/__config
miscco: I believe the `__SIZEOF_INT128__ ` macro is a very old one and I did not delete it out of pure paranoia

include/cuda/std/detail/libcxx/include/__config
```diff
@@ -1744,10 +1737,30 @@ extern "C" _LIBCUDACXX_FUNC_VIS void __sanitizer_annotate_contiguous_container(
...
+#  define _LIBCUDACXX_HAS_THREAD_API_CUDA
+#endif // __cuda_std__
+#endif // _LIBCUDACXX_HAS_THREAD_API_CUDA
+
+#ifndef _LIBCUDACXX_HAS_THREAD_API_WIN32
+#if defined(_LIBCUDACXX_COMPILER_MSVC)        \
+ && !defined(_LIBCUDACXX_HAS_THREAD_API_CUDA)
+#  define _LIBCUDACXX_HAS_THREAD_API_WIN32
+#endif
+#endif // _LIBCUDACXX_HAS_THREAD_API_WIN32
```
griwes, line 1758: Doesn't this change what things are defined, compared to the current code? It seems to me that for MSVC, we used to... define only API_WIN32 in the host TU and only API_CUDA in the device TUs, and with this new definition we'll end up defining both for both kinds of TUs. Maybe there's no difference in how these are currently used, but I don't think this is quite correct either way; the way upstream defines them below definitely assumes that only one is defined at once. (I think we were also wrong in how we defined API_EXTERNAL earlier.)

include/cuda/std/detail/libcxx/include/__config
```diff
@@ -445,487 +445,500 @@ extern "C++" {
...
+ || defined(_LIBCUDACXX_COMPILER_MSVC)                       \
+ || defined(_LIBCUDACXX_COMPILER_NVRTC)
+#define _LIBCUDACXX_UNDERLYING_TYPE(...) __underlying_type(__VA_ARGS__)
+#endif // __check_builtin(underlying_type)
+
+#if defined(_LIBCUDACXX_COMPILER_CLANG)
+
+// _LIBCUDACXX_ALTERNATE_STRING_LAYOUT is an old name for
+// _LIBCUDACXX_ABI_ALTERNATE_STRING_LAYOUT left here for backward compatibility.
+#if (defined(__APPLE__) && !defined(__i386__) && !defined(__x86_64__) &&       \
```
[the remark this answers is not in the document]
miscco: Its not even used anywhere, removing

include/cuda/std/detail/libcxx/include/__config
miscco: Yeah those changes were accidental. Good catch!

I reverted to the previous state

miscco:
I have addressed the THREAD_API issue and will give it a final CI run before merging

miscco:
I have verified that the preprocessor output for gcc / clang / nvc++ is either identical or uses available builtins that werent used before

PR #432, opened 2023-04-13
Move away from `__CUDA_ARCH__`

miscco:
We want to support nvc++ which does not know about `__CUDA_ARCH__`

so we replace it with NV_IF_TARGET and friends


miscco:
Note to reviewers, if you enable "hide whitespace" the diff is much easier to verify

wmaxey (approved):

griwes (approved):

PR #437, opened 2023-04-17
Implement <optional>

### LINKED ISSUE #169: Add `cuda::std::optional`
Since we're doing `std::variant` (#162), we probably ought to do `std::optional` as well.

griwes:
Implement the `<optional>` header, with a C++14 backport.

Resolves #169


include/cuda/std/detail/libcxx/include/optional
```diff
@@ -146,25 +159,52 @@ template<class T>
...
+#ifndef _LIBCUDACXX_HAS_NO_SPACESHIP_OPERATOR
+#include "compare"
 #endif
 
-_LIBCUDACXX_PUSH_MACROS
-#include <__undef_macros>
-
+#if !defined(_LIBCUDACXX_HAS_NO_PRAGMA_SYSTEM_HEADER)
+#  pragma GCC system_header
+#endif
```
miscco, line 205:
```suggestion
#if defined(_LIBCUDACXX_USE_PRAGMA_GCC_SYSTEM_HEADER)
#pragma GCC system_header
#endif
```

include/cuda/std/detail/libcxx/include/optional
```diff
@@ -146,25 +159,52 @@ template<class T>
...
+// standard-mandated includes
+
+// [optional.syn]
+#ifndef _LIBCUDACXX_HAS_NO_SPACESHIP_OPERATOR
+#include "compare"
 #endif
 
-_LIBCUDACXX_PUSH_MACROS
-#include <__undef_macros>
-
```
miscco, line 166:
```suggestion
#ifndef __cuda_std__
#include <__pragma_push>
#endif // __cuda_std__
```

include/cuda/std/detail/libcxx/include/optional
```diff
@@ -606,12 +657,14 @@ private:
     // LWG2756: conditionally explicit conversion from _Up
     struct _CheckOptionalArgsConstructor {
       template <class _Up>
+      _LIBCUDACXX_INLINE_VISIBILITY
       static constexpr bool __enable_implicit() {
```
miscco, line 661: My spider senses are tingling that this will break a lot of older supported compilers hard. 

If I remember correctly, I had to revert all those changes and move to inline variables / structs

miscco (commented):
This looks great, couple of nitpicks

.upstream-tests/test/std/utilities/optional/optional.monadic/and_then.pass.cpp
```diff
@@ -0,0 +1,302 @@
...
+};
+
+// check that the lambda body is not instantiated during overload resolution
+__host__ __device__
+TEST_CONSTEXPR_CXX17 void test_sfinae() {
+  cuda::std::optional<NonConst> opt{};
+  auto l = nvrtc_workaround(); // [](auto&& x) { return x.non_const(); };
+  opt.and_then(l);
+  cuda::std::move(opt).and_then(l);
+}
```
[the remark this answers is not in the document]
griwes: This *should* be equivalent though, no? What it's trying to check is if the template gets instantiated in wrong ways. The body of the templates are the same, and after desugaring they should be almost equivalent, sans the return type being deduced in the lambda case. Am I missing something here?

.upstream-tests/test/std/utilities/optional/optional.monadic/or_else.pass.cpp
```diff
@@ -0,0 +1,109 @@
...
+      assert(false);
+      return cuda::std::optional<MoveOnly>{};
+    });
+  }
+
+  return true;
+}
+
+int main(int, char**) {
+  test();
```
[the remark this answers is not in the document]
griwes: Ooooops. Good catch. I've split it because the nontrivial cases were giving me grief, but turns out I forgot to actually call that...

.upstream-tests/test/std/utilities/optional/optional.object/optional.object.assign/assign_value.pass.cpp
```diff
@@ -0,0 +1,304 @@
...
+#include <cuda/std/cassert>
+// #include <cuda/std/memory>
+
+#include "test_macros.h"
+#include "archetypes.h"
+
+using cuda::std::optional;
+
+struct ThrowAssign {
+  STATIC_MEMBER_VAR(dtor_called, unsigned);
```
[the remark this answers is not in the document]
griwes: Blergh. This is not just exception tests though, there's a bunch of tests that use it to verify whether functions have been called in non-exceptional cases.

I don't know how scallable STATIC_TEST_GLOBAL_VAR is; the entirety of archetypes.h uses these things. We should probably discuss this entire issue live and figure out a workable solution. Is there a chance that making STATIC_MEMBER_VAR just expand to `static` on nvc++ would work (even if theoretically UB per nvc++ itself)?

include/cuda/std/detail/libcxx/include/optional
griwes: Can you elaborate? I'm not seeing any failures around this in the CI or locally.

miscco: This was just something that broke tuple maybe it was just with older 11.X CTKs so that should not be an issue here

.upstream-tests/test/std/utilities/optional/optional.object/optional.object.observe/dereference.pass.cpp
```diff
@@ -0,0 +1,71 @@
...
+    optional<Y> opt{Y{}};
+    return (*opt).test();
+}
+
+int main(int, char**)
+{
+    {
+        [[maybe_unused]] optional<X> opt;
+        ASSERT_SAME_TYPE(decltype(*opt), X&);
+        LIBCPP_STATIC_ASSERT(noexcept(*opt), "");
```
miscco, line 54: I do not think we should use that macro

miscco (commented):
Went through the first 50 files. Some minor issues, will continue with  the rest tomorrow

.upstream-tests/test/support/test_macros.h
```diff
@@ -322,6 +323,10 @@ struct is_same<T, T> { enum {value = 1}; };
     static_assert((test_macros_detail::is_same<__VA_ARGS__>::value), \
                  "Types differ unexpectedly")
 
+// TODO: figure out if this makes sense
+#define TEST_THROW(...) assert(#__VA_ARGS__)
```
miscco, line 327: Should that really stay?

include/cuda/std/detail/libcxx/include/optional
```diff
@@ -236,8 +283,13 @@ struct __optional_destruct_base<_Tp, false>
...
            __engaged_(true) {}
 
+  template <class _Fp, class... _Args>
+  _LIBCUDACXX_INLINE_VISIBILITY
+  constexpr __optional_destruct_base(__optional_construct_from_invoke_tag, _Fp&& __f, _Args&&... __args)
+      : __val_(_CUDA_VSTD::invoke(_CUDA_VSTD::forward<_Fp>(__f), _CUDA_VSTD::forward<_Args>(__args)...)), __engaged_(true) {}
+
     _LIBCUDACXX_INLINE_VISIBILITY
-    void reset() noexcept
+    _LIBCUDACXX_CONSTEXPR_AFTER_CXX11 void reset() noexcept
```
miscco, line 292: That should only be constexpr with C++20 due to the destructor call

include/cuda/std/detail/libcxx/include/optional
```diff
@@ -317,24 +374,28 @@ struct __optional_storage_base : __optional_destruct_base<_Tp>
 
     template <class... _Args>
     _LIBCUDACXX_INLINE_VISIBILITY
-    void __construct(_Args&&... __args)
+    _LIBCUDACXX_CONSTEXPR_AFTER_CXX11 void __construct(_Args&&... __args)
```
miscco, line 377: Ditto, this should be 
```suggestion
    _LIBCUDACXX_CONSTEXPR_AFTER_CXX17 void __construct(_Args&&... __args)
```

include/cuda/std/detail/libcxx/include/optional
griwes: Disagree - if the type in the optional has a trivial destructor, this will work pre-20, no?

.upstream-tests/test/std/utilities/optional/optional.relops/equal.pass.cpp
```diff
@@ -0,0 +1,102 @@
...
+    assert(o1 == O2(42));
+    assert(!(O2(101) == o1));
+  }
+
+  return true;
+}
+
+int main(int, char**) {
+  test();
+#if TEST_STD_VER >= 17
```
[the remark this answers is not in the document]
griwes: Because they trade in optionals of non-trivial types, which can only be made constexpr-destructible in C++20+.

include/cuda/std/detail/libcxx/include/optional
griwes: Ditto.

.upstream-tests/test/std/utilities/optional/optional.object/optional.object.ctor/U.pass.cpp
```diff
@@ -0,0 +1,164 @@
...
+    static_assert(!cuda::std::is_constructible<O, Input, int>::value, "");
+    optional<To> opt(cuda::std::forward<Input>(in));
+    return opt && *opt == static_cast<To>(v);
+}
+
+__host__ __device__
+void test_implicit()
+{
+    {
+        static_assert(implicit_conversion<long long>(42, 42), "");
```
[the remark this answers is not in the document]
griwes: I'm fairly certain that the rest of this file provides sufficient runtime coverage; these static_assert-only cases generally just verify that the correct overloads get chosen, which doesn't really need a runtime component.

.upstream-tests/test/std/utilities/optional/optional.object/optional.object.observe/dereference.pass.cpp
griwes: Why not? I mean it's probably unnecessary here, because it's only our code that's tested by this test and so this extension is always present, and this is a macro for testing libc++ extensions, but I really don't see the need to generate churn of removing its usage.

.upstream-tests/test/std/utilities/optional/optional.object/optional.object.observe/op_arrow_const.pass.cpp
```diff
@@ -0,0 +1,75 @@
...
+        // ASSERT_NOT_NOEXCEPT(opt.operator->());
+        // FIXME: This assertion fails with GCC because it can see that
+        // (A) operator->() is constexpr, and
+        // (B) there is no path through the function that throws.
+        // It's arguable if this is the correct behavior for the noexcept
+        // operator.
+        // Regardless this function should still be noexcept(false) because
+        // it has a narrow contract.
+    }
+    {
```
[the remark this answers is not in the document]
griwes: That doesn't seem necessary, the one runtime assertion present already verifies that op arrow works, and the other cases just test how much of constexpr is propagated.

include/cuda/std/detail/libcxx/include/optional
miscco: It would, but that here is the specialization where `is_trivially_destructible` is false

include/cuda/std/detail/libcxx/include/optional
```diff
@@ -271,8 +323,13 @@ struct __optional_destruct_base<_Tp, true>
...
            __engaged_(true) {}
 
+  template <class _Fp, class... _Args>
+  _LIBCUDACXX_INLINE_VISIBILITY
+  constexpr __optional_destruct_base(__optional_construct_from_invoke_tag, _Fp&& __f, _Args&&... __args)
+      : __val_(_CUDA_VSTD::invoke(_CUDA_VSTD::forward<_Fp>(__f), _CUDA_VSTD::forward<_Args>(__args)...)), __engaged_(true) {}
+
     _LIBCUDACXX_INLINE_VISIBILITY
-    void reset() noexcept
+    _LIBCUDACXX_CONSTEXPR_AFTER_CXX11 void reset() noexcept
```
miscco, line 332: We are only targeting C++14, this should be plain `constexpr`

include/cuda/std/detail/libcxx/include/optional
miscco: This is calling placement new, that cannot be constexpr or am I missing something?

include/cuda/std/detail/libcxx/include/optional
```diff
@@ -351,9 +412,9 @@ struct __optional_storage_base : __optional_destruct_base<_Tp>
     }
 };
 
-// optional<T&> is currently required ill-formed, however it may to be in the
-// future. For this reason it has already been implemented to ensure we can
-// make the change in an ABI compatible manner.
+// optional<T&> is currently required to be ill-formed. However, it may
```
miscco, line 415: Right now I am in favor of just remmoving that one. We are not really caring about ABI and I do not really see why it would break ABI because you cannot use an optional with a reference type today

include/cuda/std/detail/libcxx/include/optional
```diff
@@ -700,76 +761,82 @@ public:
     constexpr explicit optional(_InPlaceT, _Args&&... __args)
         : __base(in_place, _CUDA_VSTD::forward<_Args>(__args)...) {}
 
-    template <class _Up, class... _Args, class = _EnableIf<
+    template <class _Up, class... _Args, class = __enable_if_t<
```
miscco, line 764: I had all kinds of funny issues sometimes with plain `__enable_if_t`

Should we move to Erics concept emulation?

That would also be slightly more readable

.upstream-tests/test/std/utilities/optional/optional.monadic/transform.pass.cpp
```diff
@@ -0,0 +1,243 @@
...
+// template<class F> constexpr auto transform(F&&) &;
+// template<class F> constexpr auto transform(F&&) &&;
+// template<class F> constexpr auto transform(F&&) const&;
+// template<class F> constexpr auto transform(F&&) const&&;
+
+#include "test_macros.h"
+#include <cuda/std/cassert>
+#include <cuda/std/optional>
+#include <cuda/std/type_traits>
+
```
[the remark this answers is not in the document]
griwes: The very first include here is of test_macros.h.

include/cuda/std/detail/libcxx/include/optional
```diff
@@ -999,37 +1058,165 @@ public:
...
+    static_assert(is_same_v<remove_cvref_t<invoke_result_t<_Func>>, optional>,
+                  "Result of f() should be the same type as this optional");
+    if (this->__engaged_)
+    {
+      return _CUDA_VSTD::move(*this);
+    }
+    return _CUDA_VSTD::forward<_Func>(__f)();
+  }
+
+    using __base::reset;
```
miscco, line 1207:
```suggestion
  using __base::reset;
```

griwes: Actually applied the reverse comment, the block of the monadic functions was indented with 2 spaces, while the rest of this file is 4.

libcxx/test/std/utilities/optional/optional.bad_optional_access/derive.pass.cpp
```diff
@@ -6,8 +6,8 @@
 //
```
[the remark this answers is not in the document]
griwes: These are actually unmodified by us, just pulled from upstream as they are there. Do we still need to do that to those files? Is it even correct to add our copyright to a thing we have not, in fact, modified?

include/cuda/std/detail/libcxx/include/optional
griwes: I haven't seen anything in the CI, so I'll leave it as is and fix if we do somehow get issues.

include/cuda/std/detail/libcxx/include/optional
```diff
@@ -700,76 +763,84 @@ public:
...
+        _LIBCUDACXX_TRAIT(is_constructible, value_type, initializer_list<_Up>&, _Args...)>
     >
     _LIBCUDACXX_INLINE_VISIBILITY
     constexpr explicit optional(in_place_t, initializer_list<_Up> __il, _Args&&... __args)
         : __base(in_place, __il, _CUDA_VSTD::forward<_Args>(__args)...) {}
 
-    template <class _Up = value_type, _EnableIf<
-        _CheckOptionalArgsCtor<_Up>::template __enable_implicit<_Up>()
+    template <class _Up = value_type, __enable_if_t<
+        _CheckOptionalArgsCtor<_Up>::template __enable_implicit<_Up>::__value
```
miscco, line 774: I believe we should remove all these complicated indirections and turn them into static constexpr variables.

But we can do that in a follow-up PR

include/cuda/std/detail/libcxx/include/optional
```diff
@@ -991,45 +1054,173 @@ public:
...
-        return _CUDA_VSTD::addressof(__x);
+    constexpr auto and_then(_Func&& __f) const&& {
+        using _Up = invoke_result_t<_Func, const value_type&&>;
+        static_assert(__is_std_optional<remove_cvref_t<_Up>>::value,
+                      "Result of f(std::move(value())) must be a specialization of std::optional");
+        if (this->__engaged_)
+        {
+            return _CUDA_VSTD::invoke(_CUDA_VSTD::forward<_Func>(__f), _CUDA_VSTD::move(this->__get()));
+        }
+      return remove_cvref_t<_Up>();
```
miscco, line 1114:
```suggestion
        return remove_cvref_t<_Up>();
```

miscco (approved):
One question about reset in the non-trivially-destructible case

Rest are nits

include/cuda/std/detail/libcxx/include/optional
```diff
@@ -236,8 +283,13 @@ struct __optional_destruct_base<_Tp, false>
...
            __engaged_(true) {}
 
+  template <class _Fp, class... _Args>
+  _LIBCUDACXX_INLINE_VISIBILITY
+  constexpr __optional_destruct_base(__optional_construct_from_invoke_tag, _Fp&& __f, _Args&&... __args)
+      : __val_(_CUDA_VSTD::invoke(_CUDA_VSTD::forward<_Fp>(__f), _CUDA_VSTD::forward<_Args>(__args)...)), __engaged_(true) {}
+
     _LIBCUDACXX_INLINE_VISIBILITY
-    void reset() noexcept
+    constexpr void reset() noexcept
```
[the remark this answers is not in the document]
griwes: This will work at constexpr in pre-C++17 if the branch isn't engaged though, right? Though maybe that's not useful enough to make it unconditionally constexpr?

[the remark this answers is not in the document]
griwes: I'll make it conditional.

PR #440, opened 2023-04-20
Rework how we handle feature test macros

miscco:
We do not want to conflict with the host standard library but also provide users with feature test macros


include/cuda/std/detail/libcxx/include/version
```diff
@@ -201,234 +201,468 @@ __cpp_lib_void_t                                        201411L <type_traits>
...
 # endif
-# define __cpp_lib_span                                 201902L
-# define __cpp_lib_string_udls                          201304L
-# define __cpp_lib_transformation_trait_aliases         201304L
-# define __cpp_lib_transparent_operators                201210L
-# define __cpp_lib_tuple_element_t                      201402L
-# define __cpp_lib_tuples_by_type                       201304L
-#endif
+# define __cuda_lib_span                                 201902L
+# define __cuda_lib_string_udls                          201304L
```
wmaxey, line 247: We don't have string (yet), so should sit down and do a pass over these and remove things we don't support in `cuda_std` mode?

wmaxey: I can see how this made things more difficult when defining the `cpp_lib` macros as there wouldn't be a value or macro set for them. I'm not a fan of special cases.

miscco: I took this up and separated them in case we do not implement something.

I believe this is not too bad, as we really want to track what we implement and what not

miscco:
Dropping as this will be handled in the new cccl repo

PR #449, opened 2023-05-09
Add specialization of `construct_at` for trivial types

miscco:
There is a performance bug with EDG, where it needlessly inserts nullptr checks after placement new, which affects nvcc prior to CTK 12.2

To work around this for older compilers, specialize construct at for trivially constructible types where we simply assign rather than going for placement new.

nvbug3692949


include/cuda/std/detail/libcxx/include/__memory/construct_at.h
```diff
@@ -69,13 +71,22 @@ _LIBCUDACXX_INLINE_VISIBILITY _LIBCUDACXX_CONSTEXPR_AFTER_CXX17 _Tp* construct_a
   return ::new (_CUDA_VSTD::__voidify(*__location)) _Tp(_CUDA_VSTD::forward<_Args>(__args)...);
 }
 
+template <class _Tp, class... _Args, class = decltype(::new(_CUDA_VSTD::declval<void*>()) _Tp(_CUDA_VSTD::declval<_Args>()...))>
+_LIBCUDACXX_INLINE_VISIBILITY _LIBCUDACXX_CONSTEXPR_AFTER_CXX17
+__enable_if_t<is_trivially_constructible_v<_Tp, _Args...>,_Tp*> construct_at(_Tp* __location, _Args&&... __args) {
```
senior-zero, line 76: User might overload `static void *operator new(std::size_t, void *ptr)` for a custom type, which is not going to be used by this specialization. This seems like a change of behavior. I suggest we check against fundamental types instead of trivially constructible ones.

senior-zero (changes requested):

include/cuda/std/detail/libcxx/include/__memory/construct_at.h
miscco: Yeah I checked and even trivial types can define custom placement new, so we need to restrict ourself to fundamental types

miscco: I reduced further down to arithmetic types, as neither void nor nullptr_t seem valid choices here

senior-zero: I think we've just lost vector types like int4, double2 etc. Do you think we can handle them somehow?

miscco: That would require creating a specialized type trait, so yes but not in this PR

senior-zero (approved):
Approve, provided that an issue will be filed to ensure addressing CUDA vector types.

jrhemstad (approved):

PR #454, opened 2023-05-12
Fix `construct_at` for union with trivial members

miscco:
When switching the active member of a union in a constexpr context we must go through `::std::construct_at`, otherwise this is technically UB.


miscco:
I believe the underlying issue is this:  https://eel.is/c++draft/class.union#general-5

We are not accessing the member of the union through a "member access expression" but we are forming a pointer to it. 🤷‍♂️

senior-zero (approved):

griwes:
This isn't urgent, and I'm not going to fix this in the optional PR for now, but this handling of the issue broke the constexpr usage of construct_at in the *libcxx* tests for optional. We'll need to fix that eventually when cleaning libc++-mode tests fully.

PR #462, opened 2023-05-18
Fix the atomic enum class test on SM60.

wmaxey:


griwes (approved):

wmaxey:
Running tests and will merge if green.

PR #467, opened 2023-05-24
Properly uglify all qualifiers in product headers

miscco:
This cleans up all plain uses of `cuda::std` in product code and replaces them by the uglified macro `_CUDA_VSTD`


miscco:
Dropping in favor of https://github.com/NVIDIA/cccl/pull/184

PR #470, opened 2023-05-25
Relax execution checks for function objects

miscco:
Previously, we would require that a function that is passed to a function object such as `cuda::std::plus` is annotated as `__host__ __device__`

However, this breaks valid use cases, such as using a `__device__` function in a `__global__` kernel.

This obviously also inhibits any compiler warnings which might catch a bug, but rejecting valid code seems the worse choice.


jrhemstad (approved):

miscco:
@jrhemstad I just realized, that we still require the predicate to be copy-constructible on device.

I am not really that concerned because stateful functors are rare on device, but do we want to also disable execution space checks for constructors?

PR #479, opened 2023-06-05
Implement first batch of <ranges> views

miscco:
This implements the first part of views for `<ranges>`

I am splitting that PR, as it is only getting bigger and bigger


miscco:
Closing as this will be handled in the new cccl repo

PR #485, opened 2023-06-12
Fix building with gcc-6 and make the test suite pass

miscco:


griwes (approved):

wmaxey:
It's overkill in some capacity to get tests passing on GCC-6. We just need the smoke tests and atomics. Someday we can special case that. I'm happy to see extra credit here though. 💯

wmaxey (approved):

PR #486, opened 2023-06-12
Fix link to std::binary_semaphore documentation

pauleonix:
Previous link points at empty page as cppreference documentation was merged with the page for std::counting_semaphore.


miscco (approved):
Thanks a lot, I verified the changes

PR #492, opened 2023-06-21
Mark tests that failed because of busy device as UNRESOLVED instead of failed

miscco:
We tried to rerun the tests, but it turned out to be still flaky. We are testing a sufficient number of configurations, so that we should be safe from accidental breakage.


miscco:
Closing because it is not necessary and we are moving to the cccl repo anyway

PR #495, opened 2023-06-29
Fix atomic_ref example

pauleonix:
Very small fix of variable name and reference vs pointer.


miscco (approved):
Thanks a lot for fixing this bug in the documentation 🎉  

We are currently in the process of moving the repository to https://github.com/NVIDIA/cccl/tree/monorepo

I will merge the branch here and then mirror it to the cccl repo.

PR #498, opened 2023-07-27
Replace diag_suppress by nv_diag_suppress in documentation

ahendriksen:
Fix docs.

nvcc version 11 warns that `diag_suppress` is deprecated. Version 12 does not support it anymore. `nv_diag_suppress` works.

Godbolt example to try with different compiler versions: https://godbolt.org/z/nGnP95n8r


miscco:
Hi, Could you please add "[skip-tests]" to your commit message

We want to avoid running extensive CI for documentation changes

ahendriksen:
Done. Moved here: https://github.com/NVIDIA/cccl/pull/281

